PR #67, opened 2022-11-25
Improved file checks

alex-silverman:
basically move the initial validity checks (for now: missing files, bad structure, and hashed) into a separate function.


kiadso:
Hi Alex, I tried uploading an LSA on this branch but the app crashed with this error message: 
`Warning: Error in writeImpl: Text to be written must be a length-one character vector 1: shiny::runApp`

PR #70, opened 2022-11-28
move all the headers outside the ObserveEvent so they can dynamically…

alex-silverman:
… show/hide the right text based on whether they uploaded a file or not.

also, comment out the headerDesktime and headerUtilization, as we aren't doing that now.


kiadso (changes requested):
Whenever I upload a file to this branch, I get an Error in the Upload Hashed CSV zip file that's red and says "Error: object 'Export' not found"

kiadso (approved):

PR #85, opened 2022-12-06
Updated color of buttons

vlopez0603:


kiadso:
Thank you Victoria! More contrast, more accessibility :)

PR #105, opened 2022-12-13
Fix plot discrepancies

alex-silverman:
added HouseholdID and Issue to unique selection for counting issues by org/project. Before we were counting people with any issue. But we want to count people-households for every issue.


kiadso (approved):
Looks great!! <3

PR #109, opened 2022-12-14
Create initial dataframes to beused for the plots so they are consist…

alex-silverman:
…ent across plots


kiadso (approved):
YES- this is great, @alex-silverman- thank you!

PR #121, opened 2022-12-21
Function-ified the metadata logging, using a dataframe, and write_csv. Also added logging of downloads

alex-silverman:
Also had to move the 00_functions to the top of server, so we could access the new logMetadata function upon app-load (i.e. site-visit)

### DIFF
--- a/server.R
+++ b/server.R
@@ -76,6 +76,8 @@ function(input, output, session) {
   })
   
   observeEvent(input$imported, {
+    source("00_functions.R", local = TRUE) # calling in HMIS-related functions that aren't in the HMIS pkg
+    
     initially_valid_zip <- zip_initially_valid()
     
     if(initially_valid_zip) {


kiadso (commented):
Hey I ran this and when I went to download the System-level workbook, R threw this:
`Warning: Error in select: object 'served_in_date_range' not found                                                 
  [No stack trace available]`
I will look at it more closely tomorrow!

kiadso (commented):
Same thing for Organization-level DQ download

alex-silverman:
I think this is a scoping problem. The 00_functions script is now being called *outside* the observeEvent for the upload file, but the dq_main_reactive() (where the code is erroring out) is defined *inside* the observeEvent. The dq_main_reactive() calls the calculate_long_stayers() function, which references the served_in_date_range dataset. And that dataset is defined in the 03_DataQuality script, which is also called within the observeEvent. But the calculate_long_stayers() function is defined in the 00_functions script. So when the code gets to that function, it can no longer see the served_in_date_range dataset. Ideally, we would move everything outside the observeEvent if it doesn't need to be there. I would guess any reactive() falls in that category. However, for now, it seems like copying the 00_functions script back inside observeEvent works.

Issue #118, opened 2022-12-19
Upload processing disconnects from server


eanders:
**Describe the bug**
When uploading a hashed CSV set, the site disconnects from the server.

**To Reproduce**
1. Go to https://hmis.abtsites.com/eva/
2. Upload a hashed CSV (zipped)
3. Wait ~30 seconds
4. See error:

<img width="1483" alt="Screenshot 2022-12-19 at 3 52 53 PM" src="https://user-images.githubusercontent.com/1346876/208523281-c13a3efb-2e6c-4b46-9ba8-d68db0c973ee.png">

**Expected behavior**
It shouldn't disconnect

**Desktop (please complete the following information):**
 - Operating System: Mac OS
 - Browser: Chrome or Safari

kiadso:
Thank you Elliott! Whenever you try this, is it always stopping on "Assessing your data quality..."? Also, once it's grayed out, does the Processing finish?

Also, did you try this at different times or was this all in one go? (If you try again now, does it still happen?)

Thanks again-

eanders:
Attempted two or three times earlier, all around the same time. Just tried again.  It appears to be stalled in the same spot.  I'll leave it open for a while and report back, but the others never re-connected, but I only gave them a couple minutes after the disconnect.

eanders:
Yep, it appears stuck

kiadso:
Ok, thank you, Elliot- I'm checking with our back-end folks to see if any R errors were thrown. If you do right-click and Inspect and Console, do you find anything interesting there?

eanders:
<img width="1547" alt="Screenshot 2022-12-20 at 2 32 51 PM" src="https://user-images.githubusercontent.com/1346876/208751197-a81788a4-5f72-449a-b317-a90ac02b59d4.png">

eccovia-csmith:
We are seeing the same issue after uploading a hashed csv. Only errors in the console are related to missing fonts as Elliot has shown in the screenshot

kiadso:
Hi eccovia!

Thanks for the confirmation- Elliot is working on sending us a sample dataset from their system so we can test this in our development environment. In the meantime, I will raise the font issue with our server team. 

I have a question for both of you- if you collapse both of the help boxes, can you see how the File Structure Analysis turned out? Like does it list any Errors behind the gray?

Thank you!

eccovia-csmith:
[...]

PR #129, opened 2022-12-22
118 upload processing disconnects from server

kiadso:
Meant to address #118 
If it doesn't, Eva still needs to handle an upload that doesn't contain all the needed files in a graceful way.


kiadso:
If you edit one of the file names so that it's wrong and then upload a correct file, you can see the updated message.

alex-silverman (approved):

PR #131, opened 2022-12-22
Validating client count daterange

alex-silverman:
added a max and also a validate() command to the client count boxes. The latter checks that start <= end.


kiadso (approved):
So helpful! Thanks!

PR #132, opened 2022-12-27
Datetime to date

kiadso:
- changes the way date types are checked in the File Structure Analysis, accounts for updated readr package. 
- reorganizes file structure analysis script to be more like the dq one, where there's an integrity_main kind of like there's a dq_main
- counts nulls in high priority columns as High Priority issues so that the file is rejected gracefully instead of the app crashing (fixes Green River's issue)
- various fixes to date formats in calculations so code math works


alex-silverman:
looks good!

PR #133, opened 2022-12-28
Add changelog alert

kiadso:
adds changelog and changelog alert


alex-silverman:
looks good and works for me!

PR #135, opened 2022-12-28
modify missing files modal dialog to explicilty state the issue being…

alex-silverman:
… missing files, and also print which files are missing.


kiadso (approved):
great idea @vlopez0603 go team

PR #146, opened 2023-01-08
Old referrals

vlopez0603:
Ensured the function only pulls referrals that have a result (and thus a result date) associated with them. Updated the language about this function (ie "old referral" to "outstanding referral") and added in code to pull from all project types.


kiadso:
I think a thing to think about here is how we want to show which Referral it is that's outstanding. It currently gives us the Enrollment EntryDate and HHID, but not the date of the Referral. An idea is to add another separate tab like we did with Overlaps that lists the date of the referral..

PR #148, opened 2023-01-10
Tab restructuring

alex-silverman:


kiadso (commented):
I'm merging this- my only question to the group is there are too many font sizes on the Home tab, but I don't know how to make it better. Thank you for the vision and implementation, @alex-silverman!

PR #156, opened 2023-01-19
added console logging of Export info, as well as metadata logging of …

alex-silverman:
…that same info. This will help us identify the user better for debugging and tracking/metadata analysis purposes


alex-silverman:
I think if we just merge with main, then when we merge dev to main, this will be a conflict and that we can just resolve by keeping!

alex-silverman:
Genelle - we may want to delete the current metadata.csv file first so that the new one will have these new column headers. Otherwise, the new lines will have like 10 columns worth of data, but there will still be just 3 column headers.

Issue #164, opened 2023-01-27
App crash mimics a session timeout


kiadso:
**Describe the bug**
Whenever at least one user navigates to Eva, they immediately see the new timeout screen. 

**Screenshots**
forthcoming

**Additional context**
working on this now

kiadso:
Turned out to not be a time-out issue, but when Eva crashes, this sequence is triggered. It was really related to Issue #154

kiadso:
Leaving open since an app crash should not mimic a session timeout.

alex-silverman:
here's the PR for this: https://github.com/abtassociates/eva/pull/165

PR #165, opened 2023-02-01
Change timeout sequence

alex-silverman:
instead of disconnecting them, we alert them (with a simple javascript pop-up box) and reset their imported file. We keep the disconnect process but change the text, so if Eva closes (which I think would only happen if it crashes), it tells them as much.


alex-silverman:
This comes from this issue: https://github.com/abtassociates/eva/issues/164

PR #166, opened 2023-02-06
Metadata analysis update

alex-silverman:


alex-silverman:
Looks like there is a server log file in there. Do we need that? If not, let's delete before we merge.

kiadso:
Yes, I agree- I just pinged @samcon211 so we can get *.log into the .gitignore file before merging. Let me know if you want to just do it on your side!

Issue #152, opened 2023-01-13
Add details to File Structure Analysis for No Nulls Allowed


kiadso:
**Is your feature request related to a problem? Please describe.**
User reported: It would help if the Guidance for the "Nulls not allowed in this column" issue gave more information, like which rows are causing the issue.

**Describe the solution you'd like**
List the ID in the description, or if it's all rows, say that.

PR #167, opened 2023-02-07
Adding structure analysis details

alex-silverman:
this branch makes two major updates:

1. addresses [issue 152](https://github.com/abtassociates/eva/issues/152) by printing which rows are affected if < 3, "All rows affected" if all rows are null, and an example of a row affected if > 3 and < all rows. Also modifies the text of this check to indicate that it could be a null value or an incompatible data type.
2. first-pass at switching the current Guidance column to a new Detail column containing the specifics of the check and creating a new Guidance column with placeholder text. (NOTE: @vlopez0603 will take a second pass at updating the new Guidance language by searching for the word Guidance in [00_integrity_checker .R](https://github.com/abtassociates/eva/blob/1511d0e606006658fba99241d947328032603367/00_integrity_checker.R)).

Continuing with the idea of including more detail about the checks, the code combines the check_column_names and check_column_counts functions into one and specifies which columns are extraneous or missing.

### DIFF
--- a/00_integrity_checker.R
+++ b/00_integrity_checker.R
@@ -503,7 +503,9 @@ integrity_main <- rbind(
   rel_to_hoh_invalid,
   nonstandard_destination,
   nonstandard_CLS
-)
+) %>%
+  mutate(Type = factor(Type, levels = c("High Priority", "Error", "Warning"))) %>%
+  arrange(Issue, Type)
 
 if(integrity_main %>% filter(Type == "High Priority") %>% nrow() > 0) {
   structural_issues <- 1
--- a/00_integrity_checker.R
+++ b/00_integrity_checker.R
@@ -28,6 +28,8 @@ files <- c(
   "YouthEducationStatus"
 )
 
+display_cols <- c("Issue", "Type", "Guidance","Detail")
+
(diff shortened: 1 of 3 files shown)


00_integrity_checker.R
```diff
@@ -144,138 +123,106 @@ check_data_types <- function(barefile, quotedfile) {
...
-              ),
-            ". The PersonalID must be unique within the Client.csv."
-          )),
-          NULL
-        ),
-        Type = if_else(DataTypeHighPriority == 1, "High Priority", "Error")) %>%
-      filter(Issue != "nothing") %>%
-      select(Issue, Type, Guidance)
-    
+          ". The PersonalID must be unique within the Client.csv."
```
kiadso, line 150: This line seems like it's maybe out of place unless the logic was changed so that we're checking whether the PersonalID is unique together with the data type logic

alex-silverman: this change was from commit e9210064. However, you can see from the diff that this line wasn't new; it was already in there from commit 85766d2 (dated 12/12/22). I just don't want to drop something that we wanted in there for a reason.

PR #175, opened 2023-02-17
Nbn outreach long stayers

kiadso:
added long stayers logic that will treat nbn projects and outreach projects differently when calculating long stayers, using the most recent Current Living SItuation record as the date to measure from.

ALSO: added a directory to help us devs have a separate set of scripts to test with. See the sandbox/mini-non-shiny-environment directory.


00_functions.R
```diff
@@ -361,23 +361,44 @@ is_hashed <- function() {
...
+  
+  cls_df <- validation %>%
+    left_join(CurrentLivingSituation %>%
+                select(CurrentLivingSitID,
+                       EnrollmentID,
+                       InformationDate), by = "EnrollmentID") %>%
+    group_by(EnrollmentID) %>%
+    slice_max(InformationDate) %>%
+    ungroup() %>%
+    select(EnrollmentID, "MaxCLSInformationDate" = InformationDate)
```
alex-silverman, line 374: SUPER small suggestion: I think we could simplify this a bit. Since we're left-joining to the served_in_date_range dataset anyway, we don't need to start with validation. :

`cls_df <- CurrentLivingSituation %>%
    group_by(EnrollmentID) %>%
    slice_max(InformationDate) %>%
    ungroup() %>%
    select(EnrollmentID, "MaxCLSInformationDate" = InformationDate)`

alex-silverman:
It looks like most of this is the sandbox directory set-up, which is great, but not super important for app functioning. Changes to server.R seem to mostly just be clean-up. I tested the app and it worked, so I think it's ready to be pushed. I did make one small suggestion to slightly simplify/speed-up the code, but not a big deal at all. It's working either way.

PR #215, opened 2023-03-19
Beds no clients (aka Zero Utilization)

kiadso:
Added new issue to the PDDE checker called Zero Utilization. Added it to the changelog and the app_checks script


07_PDDE_Checker.R
```diff
@@ -227,9 +229,46 @@ es_no_tracking_method <- Project %>%
   ) %>%
   select(all_of(PDDEcols))
 
+
+# Zero Utilization --------------------------------------------------------
+
+projects_w_beds <- Inventory %>%
+  filter(
+    BedInventory > 0 &
+      coalesce(InventoryEndDate, meta_HUDCSV_Export_End) >= meta_HUDCSV_Export_Start &
```
alex-silverman, line 238: super small thing, but we do something similar when we check whether Inventory Dates are outside the operating dates. However, in that check we do as.Date(meta_HUDCSV_Export_End). Do we want to do the same thing here?

alex-silverman (commented):
a couple small comments and a simple spacing/formatting commit. Otherwise, this seems ready to go!

07_PDDE_Checker.R
kiadso: Yes- good catch

kiadso: oh wait, I think it's meta_HUDCSV_ExportDate we do that for, since it comes in as ymd_hms

07_PDDE_Checker.R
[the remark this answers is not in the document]
kiadso: We are going to be looking at Utilization soon, so, not sure

alex-silverman, line 1: Just thinking, does it even make sense to refer to the Project ID in the Detail when the prior columns contain the Project ID and name? I ask because for the Missing Address check, we just say "This project's Address is missing", instead of specifying the Project ID, the way we do with the other PDDE checks...

alex-silverman:
Merging, but want to confirm decision regarding the question: does it even make sense to refer to the Project ID in the Detail when the prior columns contain the Project ID and name? I ask because for the Missing Address check, we just say "This project's Address is missing", instead of specifying the Project ID, the way we do with the other PDDE checks...

07_PDDE_Checker.R
kiadso: Yes that makes sense.!

PR #247, opened 2023-04-10
Metadata 04032023

samcon211:
This is actually for the metadata from today, 4/10/2023. Branch is mislabeled, my bad! It also has the updates to help fix the timeout issue showing an artificial jump in site hits.


alex-silverman:
After this, there shouldn't need to be any more PRs for the metadata, because nothing should change from week-to-week, except the metadata.csv file, right?

PR #263, opened 2023-04-24
Metadata 4/24/2023

samcon211:


alex-silverman:
I think we need to decide once and for all where to store the Metadata
analysis stuff and then updare the instructions accordingly. Option 1 is
the metadata-analysis folder, option 2 is the www/Metadata folder. I'm
inclined toward putting everything in the Metadata-analysis folder in order
to keep things separated.

Alex

Issue #80, opened 2022-12-06
"Incorrect Data Type" error is incorrect


malspaugh:
The telephone column should in fact be an integer. The field is a dependent field re: way in which aftercare was provided. It is not a phone number so should not be a string.
Incorrect Data Type | Error | In the Exit file, the Telephone column should   have a data type of string but in this file, it is integer
-- | -- | --

kiadso:
fix forthcoming- thank you :)

PR #381, opened 2023-09-27
reordered and added High Priority file structure issues, removed thre…

kiadso:
…e rows for which there was no match in the code because they were renamed/reworked or they don't exist in the new branch.


alex-silverman:
Just clarifying that there are 3 modified checks:

- #91 was removed because it's encompassed by #117
- #80 was removed because it's no longer needed
- #76 was removed because it's encompassed by #117

PR #395, opened 2023-10-09
Overlapping participations

kiadso:
added logic to catch whenever a project has two participation records (whether that's HMIS or CE) that overlap for 1 or more days. Added as high priority pdde issues


06_PDDE_Checker.R
```diff
@@ -227,10 +234,93 @@ zero_utilization <- Project0 %>%
...
-##### For later-------
-# Incompatible Funding Source and Project Type Funding Source X can only be used with Project Type Y. Project Type A can only be used with Funding Source B. (this will need a lot more detail, hold on this one)
+# For later.. -------------------------------------------------------------
+
+# Incompatible Funding Source and Project Type Funding Source X can only be used
+# with Project Type Y. Project Type A can only be used with Funding Source B.
+# (this will need a lot more detail, hold on this one)
+
+
+# Overlapping participations ----------------------------------------------
```
alex-silverman, line 244: ideally the two dataframes below would be "funcion"ified, since they're so similar, but seems good to me!

06_PDDE_Checker.R
```diff
@@ -227,10 +234,93 @@ zero_utilization <- Project0 %>%
...
+  arrange(CEParticipationStatusStartDate) %>%
+  mutate(PreviousCEParticipationID = lag(CEParticipationID),
+         PreviousCEStart = lag(CEParticipationStatusStartDate),
+         PreviousCEEnd = lag(CEParticipationStatusEndDate)) %>%
+  ungroup() %>%
+  filter(!is.na(PreviousCEParticipationID)) %>%
+  mutate(ParticipationPeriod =
+           interval(
+             CEParticipationStatusStartDate,
+             coalesce(CEParticipationStatusEndDate, meta_HUDCSV_Export_Date)),
```
alex-silverman, line 259: I thought we decided to generally use Export_End, rather than Export_Date. Is there a reason we're using the latter here?

06_PDDE_Checker.R
```diff
@@ -227,10 +234,93 @@ zero_utilization <- Project0 %>%
...
+             PreviousCEStart,
+             coalesce(PreviousCEEnd, meta_HUDCSV_Export_Date)
+           ),
+         OverlapYN = int_overlaps(ParticipationPeriod, PreviousParticipationPeriod),
+         Detail = paste(
+           "This project's first participation period goes from",
+           CEParticipationStatusStartDate,
+           "to",
+           if_else(is.na(CEParticipationStatusEndDate),
+                   "current,",
```
alex-silverman, line 271: should this be "today", or "Export End" rather than "current"?

alex-silverman (commented):
I'm fine to merge without further changes, but if you want to make the changes below before you merge, feel free.

alex-silverman (approved):

PR #399, opened 2023-10-12
Updated language in EvaChecks.csv from "Refused" to "Preferred Not To…

vlopez0603:
- Updated language in EvaChecks.csv from "Refused" to "Preferred Not To Answer."
- Also updated "refused" in helper_functions.R to "prefers not to answer."


public-resources/EvaChecks.csv
alex-silverman, line 1: Only comment I would make is that we aren't being consistent between using "Client _**prefers**_..." and "Client _**preferred**_"

alex-silverman (commented):
One small comment on tense - in one place we have **_prefers_** and in the other we have _**preferred**_

public-resources/EvaChecks.csv
vlopez0603: This is a good point. The responses in the HMIS Data Standards are in the present tense (i.e. Client Doesn't Know, Client Prefers Not To Answer), so I would be open to going back and redoing these to stick with that tense.

vlopez0603: Updated language and pushed the commit to this branch.

PR #400, opened 2023-10-13
Add pdde rrh sso w inv

kiadso:
added pdde check for times when an RRH SO project has an active inventory record that overlaps the projects operating dates and has a BedInventory saved that's greater than 0


06_PDDE_Checker.R
```diff
@@ -227,6 +227,28 @@ zero_utilization <- Project0 %>%
...
+      interval(InventoryStartDate,
+               coalesce(InventoryEndDate, meta_HUDCSV_Export_End))
+  ) %>%
+  select(InventoryID, ProjectID, InventoryActivePeriod, BedInventory) %>%
+  left_join(Project, join_by(ProjectID)) %>%
+  mutate(RRHSOyn = ProjectType == 13 & RRHSubType == 1,
+         RRHSOActivePeriod =
+           interval(OperatingStartDate,
+                    coalesce(OperatingEndDate, meta_HUDCSV_Export_End)),
+         Detail = "") %>%
```
alex-silverman, line 245: Do we not want any detail here?

hardcodes.R
```diff
@@ -121,4 +121,7 @@ issue_levels <- c("High Priority", "Error", "Warning")
 issue_display_cols <- c("Issue", "Type", "Guidance", "Detail")
 
 # EvaChecks data (contains issue, type, guidance for each check) ----------
-evachecks <- read_csv("public-resources/EvaChecks.csv", show_col_types = FALSE)
\ No newline at end of file
+evachecks <- read_csv("public-resources/EvaChecks.csv", show_col_types = FALSE)
+
+evachecks_no_dupes <- evachecks %>%
+  janitor::get_dupes(ID) %>% nrow() == 0
```
alex-silverman, line 127: is this just in case we don't deal with merge conflicts successfully or accidentally have two checks with the same ID?

In general, I think this brings up the more general question of how to catch/flag issues internally, without disrupting the app. Like this "evades" or "covers up" the issue. I think this is where testthat and stuff comes in. Like I'm thinking we should add a simple assertthat test that raises a warning in the console if there are any dupes.

I need to be much better about that sort of unit testing.

sandbox/change-sample-data/rrh-so-has-inventory.R
```diff
@@ -0,0 +1,46 @@
+# run 00-Start_here.R EXCEPT for the "Data Prep" section
```
alex-silverman, line 1: Gawd, I love these so much. thank you thank you thank you!

alex-silverman (commented):
Only one pertinent comment oin whether we want to add Detail to this new check. I can see why not; the guidance is pretty straightforward. 

And then one other more general comment about adding unit tests as we go. Like on top of handling potential bugs in the code (like evachecks_no_dups), also adding a test that actually flags the problem.

06_PDDE_Checker.R
kiadso: I thought about it, looked at it in the output workbook and couldn't think of anything that would help beyond what's there but if you think of something lmk, we can always add it later! Thanks Alex!

hardcodes.R
kiadso: YEP, that was the line of thinking, and there's definitely something missing here that needs to be expanded to be more meaningful, but I wanted to keep it in the script so we remember to look at it at least sometimes. I agree the test_that functions could really help us out here once we get those going

alex-silverman (approved):
Let's try to remember to modify the evachecks dups file to use testthat or whatever to actually flag issues. But this is good to go by me!

PR #402, opened 2023-10-17
Demote more

kiadso:
This fixes an issue reported where there are duplicate rows for certain csv files in the columns.csv file. In order to check that there weren't more duplicates, I alphabetized the data by File and ColumnNo. This helped me find a few more that were duplicated. These were all corrected and the file is now alphabetized so it's easier to navigate.

Also, this branch demotes a LOT of columns so the only ones left are mostly primary and foreign keys and highly important columns. These are the ones currently designated as high priority:

image

### Columns to consider setting back to High Priority once we are using them in analyses:

- Assessment.AssessmentID
- Assessment.AssessmentDate
- Assessment.AssessmentType
- Assessment.PrioritizationStatus
- CEParticipation.AccessPoint
- CEParticipation.ReceivesReferrals
- CurrentLivingSituation.CurrentLivingSituation
- Enrollment.EnrollmentCoC (nullable)
- Event.Event
- Event.ReferralResult
- Event.ResultDate
- Inventory.CoCCode
- ProjectCoC.CoCCode


public-resources/columns.csv
```diff
@@ -116,12 +229,27 @@ Enrollment,68,DateUpdated,datetime,0,0
 Enrollment,69,UserID,character,0,0
 Enrollment,70,DateDeleted,datetime,0,1
 Enrollment,71,ExportID,character,0,0
+Event,1,EventID,character,1,0
+Event,2,EnrollmentID,character,1,0
+Event,3,PersonalID,character,1,0
+Event,4,EventDate,date,1,0
+Event,5,Event,numeric,0,0
```
alex-silverman, line 236: This feels like maybe it should be HP? We use Event when calculating oustanding referrals (id=100). If Event is Null, we risk not flagging cases. But maybe that's okay; We'll flag if needed once they fill in Event, as long as they notice and take care of null errors.

kiadso: I agree, it's an important data element and it should be shifted back to High Priority soon. For now we have to shift it to Error per JJ.

PR #406, opened 2023-10-19
system dataframe for analysis

kiadso:
Created system_df, a dataframe to use for the Phase 2 Eva system analyses (and caught the branch up from ages ago before when we were still on FY22.)


sandbox/mini-non-shiny-environment/07_system_overview.R
```diff
@@ -0,0 +1,75 @@
+
+system_df_prep <- EnrollmentAdjust %>%
+  left_join(Project0 %>%
+              select(ProjectID, ProjectName, OrganizationID, RRHSubType),
+            by = "ProjectID") %>%
+  left_join(Organization %>% select(OrganizationID, OrganizationName),
+            by = "OrganizationID") %>%
```
alex-silverman, line 7: I think you're only joining to Organization to get OrganizationName, but Project0 already has it (line 92 of initial data prep). So let's just add OrganizationName to the Project0 join

04_initial_data_prep.R
```diff
@@ -87,7 +87,12 @@ if("ProjectTimeID" %in% colnames(ProjectsInHMIS)){
 
 # This dataset is used when we need an unduplicated concise df for project
 Project0 <<- Project %>%
```
alex-silverman, line 89: Couldn't make this comment above, so I'm putting it here.

I noticed we're only using ProjectStaging (created on line 26) once throughout the app. Then we immediately use it to create ProjectHMIS. So I think we should combine those two blocks of code, or, if we intended to use ProjectStaging instead of Project thoughout the rest of the app (like when we create Project0), then we need to do a find+replace.

04_initial_data_prep.R
```diff
@@ -87,7 +87,12 @@ if("ProjectTimeID" %in% colnames(ProjectsInHMIS)){
 
 # This dataset is used when we need an unduplicated concise df for project
 Project0 <<- Project %>% 
-  select(ProjectID, ProjectName, OrganizationID, OrganizationName, ProjectType) %>%
+  select(ProjectID,
+         ProjectName,
+         OrganizationID,
+         OrganizationName,
```
alex-silverman, line 93: If ProjectStaging merges OrganizationName into the Project dataset, then how do we have OrganizationName when we create Project0 here? I'm guessing we should be using ProjectStaging here, instead of Project. But since the app hasn't thrown an error about it (?) maybe Project already has OrganizationName? In which case, we don't need ProjectStaging...

alex-silverman (changes requested):
No rush, but I do think we should take this time to make some of the small code combining updates I suggest. Or at least talk through them.

04_initial_data_prep.R
```diff
@@ -274,6 +280,23 @@ Enrollment <- Enrollment %>%
   mutate(AgeAtEntry = age_years(DOB, EntryDate)) %>%
   select(-DOB)
 
+# to be used for system data analysis purposes. has been culled of enrollments
+# that fall outside of participation/operation date ranges.
+
+EnrollmentAdjust <- EnrollmentAdjust %>%
```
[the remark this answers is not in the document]
kiadso: addressed- thank you for pointing this out! will push edits soon

PR #413, opened 2023-11-02
removed slowness warning. We think it's only affecting Abt computers …

alex-silverman:
…due to anti-virus or DLP or firewall setup


kiadso (approved):
Thank you @alex-silverman

PR #414, opened 2023-11-02
fixing incorrect flagging of overlapping CE Participation periods

alex-silverman:
There were 2 issues: 

1. If a period's Participation End Date was missing, we set it to be Export End Date. But if that Export End date was PRIOR to a period's start date, and another period from the project ended after the Export End Date, this could cause the record to be flagged. E.g. 12/1/2019-9/30/2023 and 10/1/2023-current could be flagged if the Export End Date was 9/1/2023.
2. We were not actually filtering the dataframe to only those records that had an overlap. Rather it was showing any projects that had multiple participation records.


06_PDDE_Checker.R
```diff
@@ -277,14 +277,16 @@ overlapping_ce_participation <- CEParticipation %>%
...
            interval(
              PreviousCEStart,
-             coalesce(PreviousCEEnd, meta_HUDCSV_Export_End)
+             coalesce(PreviousCEEnd, no_end_date)
            ),
-         OverlapYN = int_overlaps(ParticipationPeriod, PreviousParticipationPeriod),
-         Detail = paste(
+         OverlapYN = int_overlaps(ParticipationPeriod, PreviousParticipationPeriod)
+  ) %>%
+  filter(OverlapYN) %>%
```
kiadso, line 288: couldn't find this in the tidy style guide, but in the rest of Eva, we keep with the "sentence" structure inside filter()s, so like OverlapYN == TRUE would align with the rest of Eva syntax

kiadso (changes requested):

06_PDDE_Checker.R
alex-silverman: okay I'll make the change.

kiadso (approved):

PR #425, opened 2023-11-07
bug causing issues

kiadso:
would like a preliminary push, as this is potentially causing a number of confusing issues


alex-silverman (approved):
acknowledging that we don't yet want to use the truncated versions of entrydate and exitdate, but eventually will for analysis

PR #426, opened 2023-11-08
Fix data type checks

alex-silverman:


kiadso (approved):
I really like this change, moving everything to problems() makes it cleaner- thanks @alex-silverman

PR #427, opened 2023-11-08
include HispanicLatinaeo and MidEastNAfrican columns in the check for…

alex-silverman:
… missing_race. Otherwise, as in issue 421, they may have one of those filled out but Eva will flag for missing race/ethnicity


kiadso (approved):

kiadso:
the failed check catches the zero records thing plus a removed "(s)" in the evachecks.csv. Not sure what branch that edit came in on, but we don't need that (s), since the guidance would apply to a single project at a time. So if we can make that edit to the snapshot that would be good @alex-silverman

PR #437, opened 2023-11-15
Date handling

kiadso:
multiple edits to the way Eva handles dates. Major changes:

- modifying ExitAdjust to use a date into the far future as the coalesce second
- modifying several coalesces to use this far future date as well 

This bypasses several issues we've had when the user runs their Export through a date in the past and Eva false flags some issues because several coalesce seconds are used in calculations where they shouldn't be.


alex-silverman (changes requested):
This looks great. Thanks for doing so much digging and work on this!

Only change I would make is getting rid of all the "keep or change" comments, so as not to confuse us down the line. Those were jjust for while you were reviewing, right? I guess maybe same with places where you flagged use of ExitAdjust in the comments?

PR #448, opened 2023-11-27
update readme to reflect that right single and right double quotation…

alex-silverman:
… marks are not allowed


kiadso (approved):
thank you @alex-silverman!

PR #451, opened 2023-11-30
had deleted total_rows

alex-silverman:


alex-silverman:
The issue was that we deleted the total_rows definition but not throughout the code

PR #455, opened 2023-12-04
453 overlapping hmis participation records

kiadso:
Added a step to the importFile function that removes any rows in any csv where the DateDeleted is not null. If there's a date in the DateDeleted column, we are not going to ascribe any meaning to it other than it is deleted data and are removing it from anything Eva will see. We do this before we check for duplicate IDs, etc.
Also updated all the sandbox scripts to recent edits.
For testing, I added an extra row to the Organization.csv and added a DateDeleted to it.


alex-silverman (approved):
looks great. Thanks!!

alex-silverman:
I didn't realize I was on this branch and made a few testing-related updates. 

- removed the wrong-file-type dataset creation check
- removed corresponding libraries
- updated renv.lock.
- added some new code within the helper-functions.R importFile function to capture if the uploaded file doesn't end with .zip. I guess it's just a way to "validate' our actual method which uses tools::file_ext to detect the file extension.

PR #472, opened 2024-01-12
471 error no inventory records for service only rrh programs

kiadso:
adds a condition to the logic that checks the RRHSubtype since RRH-SO projects shouldn't have inventory


alex-silverman (approved):
Looks good. Do we want a changelog entry?

alex-silverman:
I looked at the diffs from the failed tests, and there's just one:

11 "No Inventory Records" PDDE errors, instead of 12. This was because of the removal of Organization A | Project ID 1492 from the list of projects with this error. Given that that project is an RRH, I'm guessing this is an expected change. If you agree, can you try and update the test results or do I need to do that?

PR #481, opened 2024-02-14
Demo mode

alex-silverman:


kiadso (changes requested):
This all looks good, but I did find that if you click Demo mode, then Cancel, you get the tab popups, which is not intentional. I updated all the language to match what's in the finalized document from @alyssandrichik. Thanks!!

alex-silverman:
I reviewed checks and they seem fine

PR #483, opened 2024-02-26
make datedeleted HP. Now that we're filtering out rows where datedele…

alex-silverman:
…ted is non-null, if the column is missing, Eva will crash

### DIFF
--- a/public-resources/columns.csv
+++ b/public-resources/columns.csv
@@ -5,7 +5,7 @@ Affiliation,3,ResProjectID,character,0,0
 Affiliation,4,DateCreated,datetime,0,0
 Affiliation,5,DateUpdated,datetime,0,0
 Affiliation,6,UserID,character,0,0
-Affiliation,7,DateDeleted,datetime,0,1
+Affiliation,7,DateDeleted,datetime,1,1
 Affiliation,8,ExportID,character,0,0
 Assessment,1,AssessmentID,character,0,0
 Assessment,2,EnrollmentID,character,1,0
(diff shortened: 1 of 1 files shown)


kiadso (approved):
tested on ACE, worked well- thanks @alex-silverman

kiadso:
@alex-silverman I looked at the testing and I'm not understanding why it's failing. It says it's in the missing-files check but I see no difference in the images. There are also several other "new" images, but they're not different..

alex-silverman:
I finally may have figured out how to fix the all-files-needing-to-be-committed thing:

"[These changes mean that metadata about your file changed, however the content of your file did not.](https://www.sesync.org/resources/faqs/why-does-git-show-all-my-files-changed-when-i-didnt-change-them#:~:text=These%20changes%20mean%20that%20metadata%20about%20your%20file,one%20of%20the%20the%20following%20sets%20of%20commands.)"

I ran the system() one at the bottom and it got rid of the files in my Git window!!!

alex-silverman:
I think I figured out the issue. You can see the final commit. Basically there was a problem in the way I was restoring the Export file after the test-missing-export script ran (and before the test-missing-multiple-files script ran). I was missing the "row.names = FALSE" argument, which I have in all the other cases. As a result, the restored Export.csv file had an extra column at the beginning with no header, i.e. the index/row names column. That caused the get_col_types function to return "character(0)ccnnnnTT..." or whatever. The problem being that "character(0)" was then being fed to unzip() as a set of column types, c-h-a-r-a-c-t-e-r-(-0-). And since "h" is not a valid character type, it failed there. Not sure why it gave the 404 error it did instead of literally anything more helpful.

My only guess as to why this hadn't come up before is that something changed in the shinytest2 or unzip packages.

Anyway, I've fixed. The only differences I see now are expected - super small file size reductions due to the removed column.

PR #487, opened 2024-04-01
Demo mode - to phase2-dev

alex-silverman:
Ideally, we will update demo mode to incorporate the waterfall stuff


kiadso (approved):
thank you @alex-silverman you're the best!

alex-silverman:
@kiadso , re-requesting your review because I added commits since you did your last review.  After testing further, more things were not working. I'm actually not sure if you already saw all of those and approved them, but jic not, I wanted to re-request.

PR #500, opened 2024-04-08
Eva global fix - pre demo mode

alex-silverman:


kiadso (approved):
reviewed all changes, tested iteratively, checking environments and their objects, added changelog entry, and the unit testing we have in place caught an issue around the FSA and unhashed files, which we fixed together and which passed the second time through. THANKS @alex-silverman!

PR #503, opened 2024-04-08
Demo mode (no globals)

alex-silverman:


demo_management.R
```diff
@@ -0,0 +1,208 @@
...
+  
+  shinyjs::show('fileStructureAnalysis')
+  
+  print("Switched to demo mode!")
+  logMetadata("Switched to demo mode")
+  
+}
+
+deactivate_demo <- function() {
+  browser()
```
kiadso, line 131: remove browser()

changelog.R
```diff
@@ -1,8 +1,18 @@
 output$changelog <- renderTable({
   tribble(
     ~Date, ~Change,
+    "04-09-2024", "Moved all variables from global environment to a session
```
kiadso, line 4: Add entry: 05-02-2024, Added demo mode, where users can try out Eva even if they don't have access to an FY 2024 compliant HUD CSV Export. (or something similar)

kiadso (changes requested):
left a few comments with minor things that need attention, and I think demo mode looks great!! The only other thing not listed in this review is what we talked about regarding plots not clearing between uploads in the same session. Not a security issue, but may be initially confusing to someone who uploaded Export A, looked at System-level (or whatever), then uploaded Export B in the same session, and went back to System-level. For about 3 seconds, depending on how big Export B is, the data from Export A will be visible in the plot. (Within the same session!)

kiadso (approved):

kiadso (commented):
tested by uploading multiple sample data files and checking for data bleeding, went from demo mode on to off, downloaded all the files and spot checked that they were as expected. also ran it in shinytest and found an issue that I fixed that didn't really have anything to do with demo mode but needed attention anyway (different column names for the FSA)

PR #505, opened 2024-04-12
Fix client count download

alex-silverman:


kiadso (approved):
Ran on ACE and downloaded Client Count report, tabs were fixed. shinytest is working- thanks @alex-silverman!

PR #511, opened 2024-05-16
make sure the FSA table is displayed after an upload. We hide it when…

alex-silverman:
… demo mode is turned off so as not to leave a big blank space there. However, this means that even while the underlying data may update after an upload, the FSA table won't display. This ensures it does

### DIFF
--- a/changelog.R
+++ b/changelog.R
@@ -1,8 +1,8 @@
 output$changelog <- renderTable({
   tribble(
     ~Date, ~Change,
-    "05-02-2024", "Added demo mode, where users can try out Eva even if they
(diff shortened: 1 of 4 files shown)


kiadso (approved):
I reviewed the edits here and tested the app on ACE by uploading good_small, clicking through the tabs, downloading the outputs, viewing the tables/plots in each. Went to demo mode, did the same. Then I uploaded COHHIO data and looked through all of the outputs, plots, tables, downloads and all is good. Looks great, and it's so nice we're using renderDT() now! Thanks for finding that @alex-silverman!

PR #513, opened 2024-05-16
Non hp check - to dev

alyssandrichik:
Looping through files and rscript to identify if non-hp columns are referenced in the code


alex-silverman (approved):
love it! thank you!

PR #514, opened 2024-05-16
Non hp check to phase2-dev

alyssandrichik:
Looping through files and rscript to identify if non-hp columns are referenced in the code


alex-silverman (approved):
looks good to me! Some changes are being carried over from dev, which is fine. eventually, they need to be in here

PR #523, opened 2024-06-06
Metadata early birds

samcon211:
Updates


metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -28,13 +28,20 @@ library(scales)
 
 visits <-
   read_lines(here("metadata-analysis/metadata/site_visits.txt"))
-
```
kiadso, line 31: (This comment will apply to several of the edits in this PR.)

here() should point to your Eva directory, whatever it is named. adding your personal directory to the code will make it fail for everyone else. Run `here::here()` in your console and if it doesn't say [1] "~/sdrive_ce/eva_dz" then let me know so I can help you get your RStudio project set up correctly. If it does say that, then line 30 should work as is. See here for more info: https://here.r-lib.org/
@zabalad

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -139,17 +168,114 @@ counts_by_week %>%
...
+       y = "Count") +
+  theme_minimal() +
+  theme(legend.position = "bottom") +
+  guides(fill = guide_legend(ncol = 2))
+
+```
+
+```{r fig.width=7, fig.height=6}
+
+staging %>%
```
kiadso, line 245: this plot doesn't work for me. the filter leaves me with 0 rows

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -217,13 +389,160 @@ Eva did not begin tracking downloads until December 29, 2022.
 
 counts_by_week %>%
```
kiadso, line 390: this doesn't show correctly for me, the only values in the WeekEnding column in the filtered dataframe are both 7/22/2024 so I think the plot just doesn't know what to do with it

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -217,13 +389,160 @@ Eva did not begin tracking downloads until December 29, 2022.
...
+  theme(legend.position = "bottom") +
+  guides(color = guide_legend(ncol = 2))
+
+
+```
+
+
+```{r fig.width=10, fig.height=5}
+
+counts_category_week %>%
```
kiadso, line 413: this also doesn't show correctly for me, same thing as above I think!

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -217,13 +389,160 @@ Eva did not begin tracking downloads until December 29, 2022.
...
+
+
+```
+
+
+## Demo Mode Stats
+
+```{r fig.width=10, fig.height=5}
+
+counts_by_week %>%
```
kiadso, line 481: this one also doesn't show correctly for me

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -217,13 +389,160 @@ Eva did not begin tracking downloads until December 29, 2022.
...
+       y = "Count") +
+  theme_minimal() +
+  theme(legend.position = "bottom") +
+  guides(fill = guide_legend(ncol = 2))
+
+```
+
+```{r fig.width=10, fig.height=5}
+
+counts_by_week %>%
```
kiadso, line 536: this one shows blank for me! I'm probably doing something wrong- let me know!

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -217,13 +389,160 @@ Eva did not begin tracking downloads until December 29, 2022.
...
+  filter(str_detect(Event, "Download")) %>%
+  filter(str_detect(Event, "DEMO MODE")) %>%
```
kiadso, line 538: these filter functions back to back may be causing problems- you would generally say
`filter(str_detect(Event, "Download") | str_detect(Event, "DEMO MODE"))` 
it may be reading these like:
`filter(str_detect(Event, "Download") & str_detect(Event, "DEMO MODE"))`

alex-silverman:
@zabalad will implement changes suggested by @kiadso . This should fix the issue with the last 6 months graphs not being correct. Then we will review and merge this PR.

metadata-analysis/Metadata-Analysis.Rmd
zabalad: fixed

metadata-analysis/Metadata-Analysis.Rmd
zabalad: fixed

metadata-analysis/Metadata-Analysis.Rmd
zabalad: fixed

metadata-analysis/Metadata-Analysis.Rmd
zabalad: fixed

metadata-analysis/Metadata-Analysis.Rmd
zabalad: fixed

metadata-analysis/Metadata-Analysis.Rmd
zabalad: Seems to work fine now

metadata-analysis/Metadata-Analysis.Rmd
zabalad: Noted, thanks for the tip on using here() ! I removed the commented out file paths that point to my directory.

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -139,17 +164,114 @@ counts_by_week %>%
...
+  distinct(SessionID, Event, .keep_all = TRUE) %>%
+  mutate(Tab = case_when(
+    str_detect(Event, "tabChangelog") ~ "Changelog",
+    str_detect(Event, "tabClientCount") ~ "Client Counts",
+    str_detect(Event, "tabDQOrg") ~ "Organization DQ",
+    str_detect(Event, "tabDQSystem") ~ "System DQ",
+    str_detect(Event, "tabLocalSettings") ~ "Local Settings",
+    str_detect(Event, "tabPDDE") ~ "PDDE DQ",
+    TRUE ~ as.character(Event)), 
+    Tab = factor(Tab, levels = c("Local Settings", "Client Counts", "PDDE DQ", "System DQ", "Organization DQ", "Changelog"))) %>%
```
kiadso, line 218: I'm seeing a large bar in this plot for "NA" and the cause is here. You are creating a new column named Tab and populating it based on Event, but there are some cases not included, such as "User on systemExitDetail" (1) and "User on systemOverview" (192). Those would show that way in your column due to line 217, TRUE ~ as.character(Event). Then when you factor the column, you only list the categories you want to see and R populates any non-matches as "NA," thus the "NA" category on the plot. If you're meaning to filter out the Events "User on systemOverview" and "User on systemExitDetail" those should be added to the filter statement above. (Maybe list the ones you do want to see in the filter so you don't have to keep adding new ones you don't want to see.)

metadata-analysis/Metadata-Analysis.Rmd
```diff
@@ -139,17 +164,114 @@ counts_by_week %>%
...
+                      "Client Counts" = "#f3722c",
+                      "Organization DQ" = "#f8961e",
+                      "System DQ" = "#f9c74f",
+                      "Local Settings" = "#90be6d",
+                      "PDDE DQ" = "#43aa8b")) +
+ labs(title = "Tabs Visited in the Last 6 Months",
+       x = "Tab",
+       y = "Count") +
+  theme_minimal() +
+  theme(legend.position = "bottom") +
```
kiadso, line 234: Since the bars are each labeled on the x axis, the legend isn't needed- you can replace "bottom" with "none" if you want to remove the legend

zabalad: Made this edit

metadata-analysis/Metadata-Analysis.Rmd
zabalad: I'm not seeing this issue come up

PR #530, opened 2024-07-18
Fixed demo mode link to EvaChecks.csv

zabalad:
Fix didn't work properly last time but I double checked and it does work now :)


alex-silverman (approved):
looks good to me!

PR #532, opened 2024-07-19
Split fsa analysis2

alex-silverman:
split the FSA analysis so that the default is a simple export indicating which files have impermissible characters, but then allow them to download a more detailed report with the precise locations of the impermissble characters.


server.R
```diff
@@ -378,11 +371,23 @@ function(input, output, session) {
     }
   )
   
+  output$downloadImpermissibleCharacterDetailBtn <- renderUI({
+    req(nrow(file_structure_analysis_main()) > 0)
```
kiadso: I tried running this with a "good" file and I'm seeing the button once it successfully gets uploaded. When I click the button, it says "Site wasn't available"

kiadso (approved):
I love this, what a great speed improvement! I pushed a slight change to keep the button from showing up unless there's an impermissible character, thanks @alex-silverman

PR #533, opened 2024-07-25
changelog to reflect new push to Main, which involves fixes to Issues…

alex-silverman:
… 482, 490, 529, and formatting of the changelog itself

### DIFF
--- a/changelog.R
+++ b/changelog.R
@@ -1,284 +1,316 @@
-chglog <- tribble(
-  ~Date, ~Change,
-  "08-01-2024", "<b>New Features:</b> <br>
-    - Updated LivingSituation something something <br>
-    - Formatted changelog for readability <br>
-    
-    <b>Bug Fixes:</b><br>
-    - Added Enrollment ID and Project Type to exports (Issues 
-    <a href='https://github.com/abtassociates/eva/issues/482'>#482</a> and 
-    <a href='https://github.com/abtassociates/eva/issues/529'>529</a>),<br>
-    - Corrected the institutional living situations and temporary living situations 
-  to align to the HMIS Data Standards. Issue 
-  <a href='https://github.com/abtassociates/eva/issues/490'>#490</a>",
-  
-  "05-20-2024", "<b>New Features:</b> <br>
-    - Added demo mode, where users can try out Eva even if they
-    don't have access to an FY 2024 HMIS CSV Export.",
-  
-  "04-09-2024", "<b>Bug Fixes:</b> <br>
-    - Moved all variables from global environment to a session 
-    environment to ensure appropriate visibility. <br>
-    - Moved data frames \"Export\" and \"Project0\" from the global
-    environment to a session environment to ensure appropriate visibility.",
-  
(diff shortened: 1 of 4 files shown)


kiadso (changes requested):
I'm looking into how to modify the table so that it's searchable. I like that's it's paginated, but now you can't search the page to find a key word. I'll post if I find anything @zabalad

alex-silverman:
We will first try to convert it to data.table and see if that can easily make it searchable. If not, we will just merge the PR and at some point pull the changelog onto GitHub instead of being in Eva.

changelog.R
```diff
@@ -1,5 +1,17 @@
 chglog <- tribble(
   ~Date, ~Change,
+  "08-01-2024", "<b>New Features:</b> <br>
+    - Updated LivingSituation something something <br>
```
kiadso, line 4: be sure this approximate change log entry gets rewritten before it's released

changelog.R
```diff
@@ -266,8 +278,7 @@ chglog <- tribble(
 )
 
 
-output$changelog <- renderHtmlTableWidget({ 
-  chglog %>% 
-    addHtmlTableStyle(align="l") %>% 
-    htmlTableWidget(rnames=FALSE) 
-})
+output$changelog <- renderDataTable({
```
kiadso, line 281: outputs are kept in server.R. While this does work, I'm thinking we should keep it consistent and move it to server.R

changelog.R
```diff
@@ -266,8 +278,7 @@ chglog <- tribble(
...
+  chglog2 <- chglog
```
kiadso, line 282: Again, this works, but I am not sure why we're creating a dataframe named chglog2? It also works to just have "chglog" here

changelog.R
```diff
@@ -266,8 +278,7 @@ chglog <- tribble(
...
+}, rownames=FALSE, escape=FALSE)
```
kiadso, line 283: keeping to tidyverse styleguide, be sure you're linting your code, or typing it with spaces between operators. In RStudio, you can do cctrl-shift-a after selecting a bit of code and it will lint it for you

ui.R
```diff
@@ -855,7 +855,7 @@ dashboardPage(
             # collapsible = TRUE,
             # collapsed = TRUE,
             width = 12,
-            htmlTableWidgetOutput("changelog")
+            dataTableOutput("changelog")
```
kiadso, line 858: Yay, this is great, the new table looks great! I think the other thing we need to do here is remove "library(htmlTable)" from global.R and remove it from the renv snapshot as well. Thanks so much for making this work @zabalad! Let me or Alex know if you need help with renv

kiadso (changes requested):
the table looks so great! Just leaving a few things to address before we merge it- thanks so much for your work on this!

alex-silverman:
The thing about outputs living in scripts outside server.R kind of brings
back up the idea of shiny modules. I kind of like the idea of puttings
things that conceptually go together in the same place, in an external
script. Especially as server.R gets longer and longer

PR #536, opened 2024-07-31
Inflow outflow 3

kiadso:
this adds a working waterfall chart where all the filters work, including the summary and detail tabs.


07_system_overview.R
```diff
@@ -128,495 +310,465 @@ system_df_enrl_flags <- system_df_prep %>%
...
+      HispanicLatinaeo +
+      BlackAfAmerican == 0, 1, 0
+  ),
+  # Data quality check for exclusive group 2
+  DQRaceEthExclusive2 =
+    BILPOCExclusive2 +
+    WhiteExclusive2 +
+    RaceEthnicityNoneExclusive, # all rows should equal 1
+  # inclusive logic group 1
+  AmIndAKNativeInclusive1 = if_else(AmIndAKNative == 1, 1, 0),
```
alex-silverman, line 667: At the very least, I don't think we need these Inclusives or we could simply rename the original variables. Mutations can lower performance. I get that naming the "Inclusive" makes it clear they are the parallel versions of the Exclusive ones. But the only difference between these and the originals are that the NAs have been replaced with 0s, and I THINK their only purposes are (1) creating the Exclusive variables, and (2) applying the user-specified filters, i.e. when you later do something like `if_any(.cols = c(input$syso_race_ethnicity)`. If we simply used the original versions that have NAs in them, I think that should be fine for both purposes, since, ultimately the important thing is that if_any function, which only reads 1s. It shouldn't matter if the values are NA.

07_system_overview.R
```diff
@@ -128,495 +310,465 @@ system_df_enrl_flags <- system_df_prep %>%
...
-  ), by = .(PersonalID, in_date_range)]
+})
+
+# Enrollment-level reactive -----------------------------------------------
+
+enrollment_categories_reactive <- reactive({
   
+  # Filter enrollments by hhtype, project type, and level-of-detail inputs
+  enrollment_categories %>%
+    filter((input$syso_hh_type == "All" |
```
alex-silverman, line 723: this stuff also *seems* redunant. I know you know that, though.

07a_system_activity_plots.R
```diff
@@ -1,283 +1,264 @@
...
+      ),
+      Status = factor(
+        Status,
+        levels = c(
+          "Homeless",                          
+          "Housed",
+          "Newly Homeless",
+          "Returned from \nPermanent",
+          "Re-engaged from \nNon-Permanent",
+          "Exited to \nNon-Permanent Destination",
```
alex-silverman, line 82: i think we decided to shorten this to "Exited, Non-Permanent" to match the sankey chart and to shorten the label

07a_system_activity_plots.R
```diff
@@ -1,283 +1,264 @@
...
+          "Exited to \nPermanent Destination",
```
alex-silverman, line 83: i think we decided to shorten this to "Exited, Permanent" to match the sankey chart and to shorten the label

alex-silverman (commented):
Thank you so much for slogging through this. While I do have a number of comments about comments and condensing code, I think it all works and makes sense, so in the interest of time, I think it's fine to merge this PR and address these comments later!

alex-silverman (approved):
approving!

PR #539, opened 2024-08-08
Correct living situation groups

kiadso:
noticed an inaccuracy in at least one of the vectors and decided to review the Data Standards for confirmation that these are correct. Changed the vectors to read like the Data Standards, which feels more future-proofed, as they add more categories, we likely will not need to make any changes to these vectors.
Also removed an unused vector: all_project_types2


hardcodes.R
```diff
@@ -101,8 +100,8 @@ project_types_enrolled_homeless <- c(lh_project_types, 14)
    
 long_stayer_percentile_project_types <- c(0, 2, 3, 8, 9, 10, 12, 13)
 
-all_project_types <- c(0, 1, 2, 3, 4, 6, 8, 9, 10, 11, 12, 13, 14) # minus Other
-all_project_types2 <- c(0, 1, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12, 14)
+all_project_types <- c(0, 1, 2, 3, 4, 6, 8, 9, 10, 11, 12, 13, 14) 
+# minus Other^
```
vlopez0603, line 104: Just wanted to confirm that though we are excluding Other in this vector, that doesn't have an impact on Inflow/Outflow logic where we do reference Other.

In what scenarios do we use this vector as opposed to a vector that _does_ include Other along with all other project types?

hardcodes.R
```diff
@@ -20,9 +20,9 @@ allowed_destinations <-
   c(116, 101, 118, 215, 206, 207, 225, 204, 205, 302, 329, 314, 332, 312, 313,
     327, 422, 423, 426, 410, 435, 421, 411, 30, 17, 24, 8, 9, 99)
 
-perm_destinations <- c(422,423,426,410,435,421,411)
+perm_destinations <- c(400:499)
```
vlopez0603, line 23: Curious on your thoughts, Genelle, on whether separate vectors are needed for _destinations and _livingsituation or if having them combined into vectors like perm_situation, temp_situation, lh_situation etc. would be useful and simplify things further.

This approach uses the same terminology as the Data Standards and would rely on whatever data field is referenced in the code (3.917, 4.12, or 3.12) to determine which of the values in the list to pull. 

For example, if some code said something like "IF Enrollment.LivingSituation IN (perm_situation) THEN blah blah blah," we would expect it not to include values 422, 423, or 426 out of the range 400:499 since those three values should only be options in an HMIS for destinations.

That said, one of the things I like about having the vectors separate is just the clarity around the vector names, even if the values are the same across the vectors (i.e., perm_destinations = 400:499 and perm_livingsituation = perm_destinations).

vlopez0603 (commented):
I marked a place where it seemed like there was a potential discrepancy between two _livingsituation vectors involving value 302 (TH).

And then I had two general comments related to further vector simplification and project type Other.

hardcodes.R
kiadso: Hi @vlopez0603, the assumption behind including the entire set of permanent living situations is that the data is coming out of HMIS in the way it's collected, so a value like 422 wouldn't be represented in the Destination data element. Looking for it wouldn't cause any harm if the assumption can be made that HMIS users are not allowed to select the 422 value in the Destination drop list in their software. If we're worried about this, we could continue to list them out each individually. I have not personally seen this problem in the SSVF data I work with or any of the sample data sets we've gotten. I guess if that was happening, what would we do with that value? Filter it out or count it as missing or would we consider it permanent and notify the community? 
If we are good with assuming that there won't be 422's (as an example) in Destination data and that if there were that we'd count them as permanent, then it is efficient to use this method since as future data standards come out, we won't have additional values to add to the lists, they would already be categorized by their value as to whether they're permanent, temporary, etc.

kiadso: Also I just noticed that in the FSA, we do check for valid Prior Living Situation and Destination values, so if those values are coming in the wrong places, they are flagged.

kiadso: Hi again, :) I noticed that we don't use "temp_destinations" anywhere so I'll remove that and we can just use temp_livingsituation. 
We do use both perm_destinations and perm_livingsituation, but I think to keep things consistent, we should switch to using perm_livingsituation even though perm_destinations is such an icon, haha. It just seems cleaner and less confusing for other devs to do it that way.

hardcodes.R
kiadso: We only ever use this in the hardcodes.R script to build the grid of what elements are required to be answered by whom (see the inc_ncb_hi_required_prep data frame at the bottom.)
If we truly want all project types, we wouldn't need to filter by ProjectType at all, and in that case Other would come in with everything else. I used "all" here to mean all Project Types defined by HUD. (I'll add a comment to that effect.) Let me know if you think we should be more explicit in the vector name.

vlopez0603: I think as long as the comment details why Other is excluded, the vector name is fine as-is.

hardcodes.R
vlopez0603: Agree with making the switch to using just the _livingsituation vectors for consistency and efficiency.

vlopez0603 (approved):
Reviewed responses to my original comments, and I agree with the changes made to update the _livingsituation vectors. Reviewed the code in the first 2 files but not the scratch file since the file was deleted.

hardcodes.R
```diff
@@ -365,22 +355,23 @@ inc_ncb_hi_required <- unnest_longer(inc_ncb_hi_required_prep, ProjectType) %>%
 
 sys_comp_filter_choices1 = c(
   "Age", 
-  "Domestic Violence Status",
```
alex-silverman, line 368: We should change this in system_composition_functions.R, too

hardcodes.R
```diff
@@ -365,22 +355,23 @@ inc_ncb_hi_required <- unnest_longer(inc_ncb_hi_required_prep, ProjectType) %>%
...
+  "Domestic Violence",
   "Gender",
-   #"Homelessness Type", # Victoria, 8/15/24: Not including this for Launch
+  "Homelessness Type",
```
alex-silverman, line 360: are we including this?

kiadso: No I think this got overwritten accidentally. I'll fix it

PR #543, opened 2024-08-13
Sankey chart

alex-silverman:
Let's hold off on merging until we merge the fixing-all-served


alex-silverman:
@kiadso - note that this PR also contains Diana's changelog commits

kiadso:
Whenever I run it on Households with Children (or likely anything that causes one of the categories to be 0) the color palette changes its associations
image

alex-silverman:
this is proving trickier than I anticipated. Alluvials are hard to customize when it comes to colors.

alex-silverman:
@kiadso - I think i fixed this by overlaying a geom_rect() that constructs the Begin bars.

kiadso (approved):
This looks great! When I use the hashed current good dataset, I see a smallish sliver where the text label is illegible. I like the idea of moving the totals out to the category names like "Exited, Permanent: 138" so there's no numbers on the bars at all, but I'll stay in my lane and leave that for VL and Jesse! :)

alex-silverman:
merging this as the first draft of the PR.

PR #545, opened 2024-08-13
removed mention of LivingSituation changes since they're already ment…

kiadso:
…ioned in the Bug Fixes section. Added the split fsa entry


alex-silverman (approved):
thanks!

PR #547, opened 2024-08-19
(Don't) Turn waterfall

kiadso:
broke out inflow and outflow types on the Detail tab of the Waterfall, shortened labels a tiny bit for easier reading


07_system_overview.R
```diff
@@ -978,9 +978,9 @@ inflow_outflow_df <- reactive({
 
       OutflowTypeDetail = case_when(
         perm_dest_client == TRUE ~
-          "Exited to \nPermanent Destination",
+          "Permanent\nDestination",
```
alex-silverman, line 981: I thought we decided to relabel this: "Exited, Permanent", to be consistent with the Sankey chart. Any reason you can think of NOT to do this?

07_system_overview.R
```diff
@@ -978,9 +978,9 @@ inflow_outflow_df <- reactive({
...
         temp_dest_client == TRUE ~
-          "Exited to \nNon-Permanent Destination",
+          "Non-Permanent\nDestination",
```
alex-silverman, line 983: And this would be "Exited, Non-Permanent"

alex-silverman:
@kiadso It also looks like the Detail Box is showing "All" for Household Type and Level of Detail. Did we decide to only show those if they're not "All"? If so, no need to change, since, I'm hoping, my most recent Fixing All Served commit will achieve this. If not, let me know, because I may have misunderstood your comment on the Fixing All Served PR.

alex-silverman:
> > @kiadso It also looks like the Detail Box is showing "All" for Household Type and Level of Detail. Did we decide to only show those if they're not "All"? If so, no need to change, since, I'm hoping, my most recent Fixing All Served commit will achieve this. If not, let me know, because I may have misunderstood your comment on the Fixing All Served PR.
> 
> I created this branch before those fixes were merged, so we'll just have to be sure there's no conflicts once we merge this

Should be fine either way, but you can always merge phase2-dev into a feature branch to bring in those updates.

kiadso:
Hi @alex-silverman thanks for documenting all the issues- I think I addressed them all. Thanks!

07a_system_activity_plots.R
```diff
@@ -149,6 +170,7 @@ system_activity_prep_summary <- reactive({
   inflow %>%
     full_join(outflow, join_by(Status, values, Time)) %>%
     full_join(frame_summary, join_by(Status, Time)) %>%
+    slice(1:3, 6, 5, 4) %>%
```
alex-silverman, line 173: The purpose of this is to sort the bars, right? Ideally, we'd order the bars using factors or sorting by name, so it's clear what's happening. Second best is e a comment explaining the purpose

alex-silverman (approved):
Only change is that Homeless should be above Housed but be light gray, whereas Housed should be dark gray.

PR #549, opened 2024-08-20
make columns HP if we reference them explicitly. 

alex-silverman:
This uses the new script Alyssa created to identify these columns.


kiadso (approved):
Thank you @alyssandrichik !

PR #552, opened 2024-08-22
Gender single select

kiadso:
changing gender to single-select as discussed

### DIFF
--- a/07_system_overview.R
+++ b/07_system_overview.R
@@ -701,8 +701,8 @@ client_categories_reactive <- reactive({
     mutate(All = 1) %>%
     filter(
       AgeCategory %in% input$syso_age &
-        if_any(.cols = c(input$syso_gender), ~ .x == 1) &
-        if_any(.cols = c(input$syso_race_ethnicity), ~ .x == 1) &
+        !!sym(input$syso_gender) == 1 &
+        !!sym(input$syso_race_ethnicity) == 1 &
         ((input$syso_spec_pops == "Veteran" &
             VeteranStatus == 1) |
            (input$syso_spec_pops == "NonVeteran" &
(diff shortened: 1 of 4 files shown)


alex-silverman (changes requested):
Haven't had a chance to fully test, but I would think we'd need to update the server.R script, where we updatePickerInput: https://github.com/abtassociates/eva/blob/b16ba7a67936532e0be7414672d52f89af7f88e0/server.R#L964.

It may also be worth changing line 704 in 07_system_overview.R. Even if it works, it will be clearer to treat it as a single-select, rather than a multi-select:
https://github.com/abtassociates/eva/blob/b16ba7a67936532e0be7414672d52f89af7f88e0/07_system_overview.R#L704

Like maybe something as simple as:
`input$syso_gender == 1 &`

kiadso:
Thanks for catching the updatePicker! I removed multi-select-related code from there, as you suggested. The if_any() needs to stay like it is because the if_any is giving column names to check. the input$syso_gender is never 1, it's always something like "All" or "ManExclusive". It works just like the RaceEthnicity one, which is also a single-select.

kiadso:
@alex-silverman when I tested it with input$syso_gender == 1, it failed because it's looking directly at the input. R has to know you're asking it to look at a column name or at the literal input and there's no distinction made with input$syso_gender == 1. if_any() is like the across() but for filters, so by using the .cols argument of that function, it allows us to distinguish the actual input of "ManExclusive" from the value of the column named ManExclusive. https://www.tidyverse.org/blog/2021/02/dplyr-1-0-4-if-any/

alex-silverman:
@kiadso - sorry, you're right that `input$syso_gender == 1` won't work as-is. However, ``!!sym(input$syso_gender) == 1` works. This is because it first evaluates `input$syso_gender`, then looks for a variable with that name and checks if it's equal to 1. This means the code doesn't have to check if all those columns == 1, but rather just the column corresponding to the one the user selected in `inut$syso_gender`.

I see why this is less intuitive, though, since it looks like the code is checking whether input$syso_gender == 1. And I doubt it will bring us much of an efficiency gain, so I'm happy to leave as `if_any`.

alex-silverman (approved):
This looks good to me!

PR #554, opened 2024-08-26
Sankey chart fixes

alex-silverman:


kiadso (changes requested):
- the annotation gets clipped if you select Household Type "Households without children"
- same problem as described in the all-served chart with tacking "Households" onto the ends of the Household Types that already have the word Household in it
- same problem as described in the all-served chart with detail box including Level of Detail and HH Type
- I'm assuming after merging w/ phase2-dev, we'll see the single select Gender droplist and the "All Household Types" and "All People" in the HHType and Level of Detail droplists but just naming it in case this has been merged back and we're missing something

alex-silverman:
> * the annotation gets clipped if you select Household Type "Households without children"
> * same problem as described in the all-served chart with tacking "Households" onto the ends of the Household Types that already have the word Household in it
> * same problem as described in the all-served chart with detail box including Level of Detail and HH Type
> * I'm assuming after merging w/ phase2-dev, we'll see the single select Gender droplist and the "All Household Types" and "All People" in the HHType and Level of Detail droplists but just naming it in case this has been merged back and we're missing something

1 - I think the key is to increase the width of the label. Right now, it's 30, but I think we can up it to at least 40
2 - I realize that the problem in the label code I lifted from your Annotation branch is this line:
```
str_detect(input$syso_hh_type, "Household") == FALSE...
```
input$syso_hh_type might be "AC" or "AO", rather than the name of that being Households with Children or Households without Children, so this will always be FALSE. It should be:
```
str_detect(getNameByValue(syso_hh_types, input$syso_hh_type), "Household") == FALSE...
```
3 - I see in the Annotation Waterfall PR you noted that we decided to remove Level of Detail since it's captured in the total.
4 - I merged in Phase2 Dev so those changes should be reflected in this branch

PR #555, opened 2024-08-26
Annotation waterfall

kiadso:
adding total and difference.
original request to list household counts rescinded/delayed due to juice/squeeze reasons (VL/GD)


07a_system_activity_plots.R
```diff
@@ -189,75 +190,107 @@ renderSystemPlot <- function(id) {
...
+        size = 5,
+        inherit.aes = FALSE
+      ) +
+      annotate(
+        geom = "text",
+        x = mid_plot,
+        y = max(df$yend) * 1.1,
+        label = paste0(
+          "Total ",
+          getNameByValue(syso_level_of_detail, input$syso_level_of_detail),
```
alex-silverman, line 266: A couple thoughts here:

1. Do you know how I can use `annotate()` in the All Served chart, since the axes aren't continuous, so I can't set the Y value in the same way?
2. When Level of Detail = "All" won't this say "Total All"? That's what happened to me when I tried this method. I think we may need this: 
```
label = paste0(
        "Total ",
        if_else(
          getNameByValue(syso_level_of_detail, input$syso_level_of_detail) == "All",
          "People",
          getNameByValue(syso_level_of_detail, input$syso_level_of_detail)
        ),
        ": ",
        total_clients,
      ),
      size = 16
    ) 
```
I think we should set the size to be larger than the default, too. I used `size=16`

alex-silverman (changes requested):
It's a little hard to see exactly what changed, but if the only thing you added was the annotation, then my two comments are: 
1. increase the font size to 16.
2. change to display "Total People" instead of "Total All People" when Level of Detail == "All People". 

Also, I think we changed the "All" option in "Level of Detail" to be "All", not "All People". That's what it is in phase2-dev.

07a_system_activity_plots.R
```diff
@@ -189,75 +190,139 @@ renderSystemPlot <- function(id) {
...
+      filter(InflowOutflow == "Inflow") %>%
+      pull(values) %>%
+      sum()
+    
+    inflow_to_outflow <- df %>%
+      filter(Status %in% c("Inflow", "Outflow")) %>%
+      pull(values) %>%
+      sum()
+    
+    if(total_clients < 11){
```
alex-silverman, line 210: I would use `validate()` instead, to be more intuitive and consistent with how we validate no data in our charts and tables. Something like:

```
validate(
      need(
        total_clients >= 10,
        message = "The data has been filtered to less than 11 records, therefore\n
          the plot will not be displayed for privacy purposes."
      )
    )
```

07a_system_activity_plots.R
```diff
@@ -303,11 +368,6 @@ syso_detailBox <- reactive({
     
     ReportStart(), " to ", ReportEnd(), br(),
     
-    if (getNameByValue(syso_hh_types, input$syso_hh_type) != "All People")
```
alex-silverman, line 306: why did these get removed?

07a_system_activity_plots.R
```diff
@@ -189,75 +190,139 @@ renderSystemPlot <- function(id) {
...
     num_segments <- 20
     segment_size <- get_segment_size(s/num_segments)
-
+    
+    total_clients <- df %>%
+      filter(InflowOutflow == "Inflow") %>%
+      pull(values) %>%
+      sum()
+    
+    inflow_to_outflow <- df %>%
```
alex-silverman, line 205: NBD, but I would move this to below the validate, or even directly in the chart where you're printing "Total Change". That way, we don't have to run this code if we don't need to, i.e. if there's not enough data to show anyway.

alex-silverman (changes requested):
It's a little hard to tell exactly what you changed because the entire ggplot has been indented. This may become more clear if you use the `validate()` approach and don't have to nest and tab the ggplot within an if/else. There's also a conflict

07a_system_activity_plots.R
```diff
@@ -189,75 +190,139 @@ renderSystemPlot <- function(id) {
...
+        # alpha = .7,
+        size = 5,
+        inherit.aes = FALSE
+      ) +
+      annotate(
+        geom = "text",
+        x = mid_plot,
+        y = max(df$yend) * 1.1,
+        size = 16/.pt,
+        label = paste0(
```
alex-silverman, line 277: This is great. I think we should turn the label into a function, since we'll likely reuse it in the other charts. I add this at the bottom of helper_functions.R:

```
sys_total_count_line <- function(total_count) {
  return(str_wrap(
    paste0(
      "Total ",
      case_when(
        input$syso_level_of_detail == "All" ~ "People",
        input$syso_level_of_detail == "HoHsOnly" ~ "Heads of Household",
        TRUE ~
          getNameByValue(syso_level_of_detail, input$syso_level_of_detail)
      ),
      case_when(
        input$syso_hh_type == "All" ~ "",
        str_detect(input$syso_hh_type, "Household") == FALSE ~
          paste(" in", getNameByValue(syso_hh_types, input$syso_hh_type),
                " Households"),
        TRUE ~
          paste(" in", getNameByValue(syso_hh_types, input$syso_hh_type))
      ),
      ": ",
      total_count
    ),
    width = 30
  ))
}
```

So you'd have:
```
annotate(
  geom = "text",
  x = mid_plot,
  y = max(df$yend) * 1.1,
  size = 16 / .pt,
  label = paste0(
    sys_total_count_line(total_clients),
    "\nTotal Change: ",
    inflow_to_outflow
  )
)
'''

File: 07a_system_activity_plots.R [hunk 6]
```diff
@@ -206,81 +206,149 @@ renderSystemPlot <- function(id) {
...
+                "Total ",
+                case_when(
+                  input$syso_level_of_detail == "All" ~ "People",
+                  input$syso_level_of_detail == "HoHsOnly" ~ "Heads of Household",
+                  TRUE ~
+                    getNameByValue(syso_level_of_detail, input$syso_level_of_detail)
+                ),
+                case_when(
+                  input$syso_hh_type == "All" ~ "",
+                  str_detect(input$syso_hh_type, "Household") == FALSE ~
```
alex-silverman, line 308: This will be reflected in the new sys_total_count_line function I wrote, but I think this line needs to change to:

`str_detect(getNameByValue(syso_hh_types, input$syso_hh_type), "Household") == FALSE...`

Otherwise, input$syso_hh_type might be "AC" or "AO", rather than the _name_ of that being Households with Children or Households without Children

kiadso: When I tested it, it returned the names of the values since they're named lists. When you run this in a browser, are you seeing the values instead of the names?

07a_system_activity_plots.R
kiadso: 1. we talked about earlier :)
2. that would return "Total People" if the input level of detail is "All", otherwise it would return "Total" plus whatever the level of detail is. In the browser I was testing in, these aligned, at least at the time of this writing, maybe some code was merged in that's causing an issue. I'll take another look!
Thanks @alex-silverman!

07a_system_activity_plots.R
kiadso: Done! Great idea, thank you @alex-silverman

07a_system_activity_plots.R
kiadso: Update: I can't recreate this issue, let me know if you're not getting the same as me

07a_system_activity_plots.R
kiadso: Nice- done, thank you!

07a_system_activity_plots.R
kiadso: can't recreate this, let me know if you're seeing an issue on your side

07a_system_activity_plots.R
kiadso: Update to my update: I DID find this and corrected it like you suggested- THANKS!

alex-silverman (approved):
Just made a small language change to the validation message. Otherwise, this looks great! Thanks for finding the annotation thing!

PR #556, opened 2024-08-26
Fixing all served

alex-silverman:
Some fixes to the All Served chart based on feedback from Jesse and Victoria (https://abtassoc.sharepoint.com/sites/21788CEToolkit/Shared%20Documents/General/Eva%20System%20Overview%20Feedback%20Tracker.xlsx?web=1)


kiadso:
- I'm noticing that when the whole row is nothing but small n's (and there's more than one of them) there's stars for that row when maybe the intent is to give a total? (Maybe you have it like they wanted, but just posing the question if those need to be obscured since as long as the total is greater than 2, we couldn't tell what the small n's were.
- If I uncheck the "All Races/Ethnicities" checkbox, I get a red error onscreen that says: "Error: object 'v_cats1_labels' not found'
- If I have Domestic Violence selected with any other checkbox, nothing loads
- If I have Veteran Status and anything else selected, I get a red onscreen error: 'Error: [object Object]'
- I wonder if the Race/Eth selectors should be radio buttons so it's clear why you can't select them both at the same time
- I was able to select Gender, All Races, and Veteran Status checkboxes at the same time
- If I select Gender and All Races (regardless of Methodology), (using the good-small sample dataset) I see some cells that have nothing there at all, no stars, no numbers. Same thing if I select Age and All Races (Methodology 2 only).
- In the annotation (or ggtitle), "Households without children" is having "Households" tacked onto the end of it which doesn't make sense- I think in the annotation code I had it search the value for the word Household first and if it didn't find it, then tack "Households" onto the end. So you would see "Children Only Households" and "Parenting Youth Households" but not "Households without children Households".
- The Level of Detail and Household Type is still in the detail box (tho this is fixed in phase2-dev, I think, so maybe it doesn't matter)
- If I switch to a different methodology, we don't start with pre-selected checkboxes like we do when the chart initially loads
- If I select Gender and All Races (good-small dataset) I see a few rows with small-n (stars) in the Total column, but no corresponding cells that indicate where the small n's are.

kiadso (changes requested):
it's looking great, I listed a number of things I noticed in a comment- let me know if you have questions!

alex-silverman:
> * I'm noticing that when the whole row is nothing but small n's (and there's more than one of them) there's stars for that row when maybe the intent is to give a total? (Maybe you have it like they wanted, but just posing the question if those need to be obscured since as long as the total is greater than 2, we couldn't tell what the small n's were.
> * If I uncheck the "All Races/Ethnicities" checkbox, I get a red error onscreen that says: "Error: object 'v_cats1_labels' not found'
> * If I have Domestic Violence selected with any other checkbox, nothing loads
> * If I have Veteran Status and anything else selected, I get a red onscreen error: 'Error: [object Object]'
> * I wonder if the Race/Eth selectors should be radio buttons so it's clear why you can't select them both at the same time
> * I was able to select Gender, All Races, and Veteran Status checkboxes at the same time
> * If I select Gender and All Races (regardless of Methodology), (using the good-small sample dataset) I see some cells that have nothing there at all, no stars, no numbers. Same thing if I select Age and All Races (Methodology 2 only).
> * In the annotation (or ggtitle), "Households without children" is having "Households" tacked onto the end of it which doesn't make sense- I think in the annotation code I had it search the value for the word Household first and if it didn't find it, then tack "Households" onto the end. So you would see "Children Only Households" and "Parenting Youth Households" but not "Households without children Households".
> * The Level of Detail and Household Type is still in the detail box (tho this is fixed in phase2-dev, I think, so maybe it doesn't matter)
> * If I switch to a different methodology, we don't start with pre-selected checkboxes like we do when the chart initially loads
> * If I select Gender and All Races (good-small dataset) I see a few rows with small-n (stars) in the Total column, but no corresponding cells that indicate where the small n's are.

2. Addressed! I hadn't defined v_cats1_labels in the right place.
3. Fixed the Domestic Violence thing. I changed which dataset the All Served pulls from and had only selected PersonalID from the enrollment_categories dataframe, but I also need DomesticViolenceCategory.
4. Fixed the issue with Veteran Status. I'd been assuming the values were "Veteran" and "Non-Veteran", rather than 1 and 0.
5. Good point on the Race/Eth selectors being radio buttons. I'm going to leave as checkboxes for now, because I think making them radio buttons would require some fiddling with the UI to make that look good/be intuitive. For now, I've tried using disabling to make it clear, but radio buttons would be clearer
6. Couldn't replicate being able to select 3 things at once. Can you show me? 
7. I think this is desired. If the cell is 0, then it's blank. This makes the buckets with values stand out more.
8. I've fixed the annotation now
9. I removed level of Detail and Household Type from the detail box
10. I think that's okay that if we switch to a different methodology, nothing is preselected. The reason for preselecting was to make it more clear to the user how the chart worked via selecting 2 things.
11. good catch. it was a simple case of missing na.rm = TRUE in a sum().

PR #561, opened 2024-09-04
Waterfall download and Prep Summary code simplification

alex-silverman:


kiadso:
The Inflow and Outflow totals are showing like this in the download: 
image

alex-silverman:
> The Inflow and Outflow totals are showing like this in the download: image

@kiadso Looks like this was the result of the new InflowOutflowSummary variable I added. I didn't update the export accordingly. Should be fixed now. Just tested on demo mode successfully.

alex-silverman:
merging now. can address further issues later.

PR #562, opened 2024-09-04
Hh labels

kiadso:
changing household labels to align w/ Stella per request, increasing plot height, and adding space where needed. 

A problem I'm hoping @alex-silverman can help with is the height of the plot is better but it spills out of the box it's in and I'm not sure how to fix that

### DIFF
--- a/hardcodes.R
+++ b/hardcodes.R
@@ -133,11 +133,11 @@ issue_display_cols <- c("Issue", "Type", "Guidance", "Detail")
 
 syso_hh_types <- list(
   "All Household Types" = "All",
-  "Households without children" = "AO",
-  "Households with children" = "AC",
-  "Children Only" = "CO",
-  "Youth and Young Adults" = "YYA",
-  "Parenting Youth" = "PY"
+  "- Youth and Young Adult" = "YYA",
+  "Adult Only" = "AO",
+  "Adult Child" = "AC",
+  "- Parenting Youth" = "PY",
+  "Children Only" = "CO"
 )
 
 syso_level_of_detail <- list(
@@ -147,7 +147,7 @@ syso_level_of_detail <- list(
 )
 
 syso_project_types <- list(
-  "All" = "All",
+  "All Project Types" = "All",
   "Residential" = "Residential",
   "Non-residential" = "NonResidential"
 )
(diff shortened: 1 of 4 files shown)


helper_functions.R
```diff
@@ -408,12 +408,13 @@ sys_total_count_display <- function(total_count) {
       ),
       case_when(
         input$syso_hh_type == "All" ~ "",
-        str_detect(getNameByValue(syso_hh_types, input$syso_hh_type), "Household") == FALSE ~
-          paste(
-            " in",
-            getNameByValue(syso_hh_types, input$syso_hh_type),
-            " Households"
-          ),
+        str_detect(
```
alex-silverman, line 411: @kiadso won't this ALWAYS be false because none of the HH types contain the word "Household" anymore?

hardcodes.R
```diff
@@ -133,11 +133,11 @@ issue_display_cols <- c("Issue", "Type", "Guidance", "Detail")
 
 syso_hh_types <- list(
   "All Household Types" = "All",
-  "Households without children" = "AO",
-  "Households with children" = "AC",
-  "Children Only" = "CO",
-  "Youth and Young Adults" = "YYA",
-  "Parenting Youth" = "PY"
+  "- Youth and Young Adult" = "YYA",
```
alex-silverman, line 136: what's with the dashes?

alex-silverman (changes requested):

hardcodes.R
kiadso: Victoria asked that they be indented like this so it's clear that this is a subset of the non-youth-related HH Types.

helper_functions.R
kiadso: yes, I just never know if they're going to change the labels again later, but yes for now

alex-silverman: I'd rather keep the code clean and reflecting the current decisions. If we change the labels, it should be pretty obvious  that the Total line looks weird. We could also always add some sort of unit test if you're worried we might miss that.

hardcodes.R
alex-silverman: not sure i fully understand:

image

What is "Youth and Young Adult" a subset of? It looks "nested" under "All Household Types", but then all HH Types are subsets of "All Household Types".

kiadso: yes it's a subset of All Household Types- here's the piece I'm trying to match (minus the Veteran-related ones) 
image

PR #570, opened 2024-09-05
-Minor updates filter display labels

vlopez0603:
- Capitalized "All Ages" in age filter
- Updated to "All People" from "All" in LOD filter
- Updated to "Child Only" from "Children Only" in HH Type filter
- Updated to "All Populations" from "None" in Spec Pop filter
- Updated DV labels in Spec Pop filter to include "Reported" based on 9/5 feedback from Jesse and Brooke


alex-silverman (approved):
looks good!

PR #574, opened 2024-09-06
Debugging all served

alex-silverman:
- simplified system composition selection choices to just one list with JS to modify the Grouped/Hispanic-Focused label based on methodology type
- Don't ignore the disabling of the selections on init, since the app defaults to selecting 2 options, so all other options should be disabled by default


kiadso (changes requested):
just need to fix the 1var code so it's ok to select only Veterans, Age, or DV without getting an error. otherwise looks good!

alex-silverman:
> just need to fix the 1var code so it's ok to select only Veterans, Age, or DV without getting an error. otherwise looks good!

Fixed with the latest commit (https://github.com/abtassociates/eva/pull/574/commits/4b5de69b7996b16abc9bff97b3ae6d78ca1159ca) which fixes a bug where I was trying to select the variable by the checkbox selection name (e.g. "Age"), rather than the underlying variable name (e.g. "AgeCategory").

PR #576, opened 2024-09-06
move DomesticViolenceCategory definition and filtering from enrollment to client-level

alex-silverman:


kiadso (approved):
pushed tiny bit of linting, otherwise math looks great- thank you for this @alex-silverman!

PR #578, opened 2024-09-06
Adding other sys overview downloads

alex-silverman:


kiadso (changes requested):
- the waterfall summary excel download works great
- the waterfall detail excel download works great
- the waterfall ppt does not download a ppt for the summary or detail. in the browser downloads section, it looks like this: 
image
- the all-served excel download doesn't seem to do anything unless I have more than one selection checked. Like if I only have 1 selection checked and click Download, it does nothing and doesn't tell me I have to check another selection. (Not sure what's expected.)
- the all-served ppt download looks the same as the other ppt download: 
image

Here is some of the console when I tried running the ppt: 
image

kiadso (approved):
thanks @alex-silverman !
The ppt's are SO CUTE! merging now

PR #579, opened 2024-09-09
add conditions that were dropped in the conversion from data.frame  data.table

alex-silverman:
for hh_adjustments


kiadso (approved):
thank you @alex-silverman ! I had to add some small logic changes and kind of slowed the processing times down with my dplyr code, but I'd like to do that in a separate branch after testing is integrated into phase2-dev (as discussed) thanks again!!

PR #582, opened 2024-09-09
More text updates

vlopez0603:
- Minor text updates to inclusive gender labels to match what's in Data Elements doc
- Updated format of date range displayed in chart subheaders to match the date format in the page subheader
- In Race/Ethnicity dropdown menu, removed "All" subsection so that "All Races/Ethnicities" is visually at a different level than the other options.


alex-silverman (approved):
looks good to me. I'd already changed the date labelling in another branch, but this should take care of it, too!

PR #586, opened 2024-09-12
modifying HouseholdType logic to have YYA include PY + UY + CO.

alex-silverman:
changing what WAS "YYA" to "UY" (unaccompanied youth).
And now YYA in the filter selection includes PY + UY + CO.
This also simplifies the filter code around HouseholdType


kiadso (approved):
looks great- thank you so much for going through it w/ me @alex-silverman!
I only did some linting and updated the comment

PR #587, opened 2024-09-13
Nbn inactives

kiadso:
Include NbN enrollments in the Inactive category if they didn't have any NbN stays in the 15 days prior to the report end.

Also fixed something with the Race/Ethnicity detailBox that wasn't working right

Also removed lines on waterfall


07_system_overview.R
```diff
@@ -983,7 +996,7 @@ universe <- reactive({
             
             # nbn shelter
             (ProjectType == es_nbn_project_type &
-               (in_date_range == TRUE | NbN15DaysAfter == TRUE)) |
+               (in_date_range == TRUE | NbN15DaysAfterReportEnd == TRUE)) |
```
alex-silverman, line 999: for consistency, I'd use `NbN15DaysAfterReportEnd == 1` instead of `NbN15DaysAfterReportEnd == TRUE`, since prior to this we are setting NbN15DaysAfterReportEnd to 1s and 0s, not TRUE and FALSE. But I know they resolve to the same thing, so nbd.

alex-silverman (approved):
one small comment on line 999 of 07_system_overview.R on consistent values for NbN15DaysAfterReportEnd, but nbd if no change made.

07_system_overview.R
kiadso: In this branch, I'm showing those are resolving to TRUE and FALSE not 1 and 0 so we'll leave it for now- thanks!

PR #588, opened 2024-09-15
In app text and label updates

vlopez0603:
- Removed "unknown" as a selectable option in the exclusive gender dropdown menu
- Updated r/e labels to be more consistent with capitalization and alphabetical order
- Updated chart name from "Composition of All Served in Period" to "Composition of All Served"
- Fixed typo in Hawaiian labels
- Updated text in exlusive/inclusive methodology labels
- Updated Interpretation Tips tables under chart Instructions subtabs
- Updated info subtab text for All Served chart


alex-silverman (approved):
Looks good to me! Thanks for the thorough find+replace! We may just want to do one more find+replace after we merge the other PRs in.

PR #591, opened 2024-09-18
Issue 477 data not collected

zabalad:
Recategorized "Data Not Collected" (99) responses to show up as a warning rather than as an error, and renamed the warnings in EvaChecks.csv to show up as "Unknown" instead of "Doesn't know/Prefers not to answer." This should be reflected in the Data Quality System-level charts as well as the exports.


zabalad:
A couple of things we should probably address:

1.) It looks like  months or times homeless is being captured twice, once by itself and once in combination with living situation
image

2) For Relationship to HOH, nulls are not a valid entry for this variable, the only permissible missing value is 99 (data not collected). With this being the case, should DNC values for relationship to HOH be treated as an error, or as a warning?

kiadso (commented):
This PR is approved, thank you @zabalad

PR #597, opened 2024-09-19
The text continues

vlopez0603:
- Added in remaining entries to glossary page
  - Enabled highlighting for search function on glossary page
- Added pop up message for System Performance Overview page when in Demo mode
- Made System Performance Overview page standalone instead of being a nested subtab
  - Updated tabName to System Overview page to be in line with all other tabNames
- Updated text for System Performance download buttons
- Changed name of "Composition of All Served" chart to "System Demographics" based on feedback from Stephanie
  - Updated image export for this chart so that it's the ReportStart and ReportEnd dates that are referenced, not the CSV export dates
- More in-app text changes to Instructions boxes throughout the app

### DIFF
--- a/hardcodes.R
+++ b/hardcodes.R
@@ -260,7 +260,7 @@ syso_race_ethnicity_excl <- list(
 
 syso_veteran_pops <- list(
   "Veteran" = "Veteran",
-  "Non-Veteran/Unknown" = "NonVeteran"
+  "Non-Veteran" = "NonVeteran"
 )
 
 syso_dv_pops <- list(
--- a/hardcodes.R
+++ b/hardcodes.R
@@ -260,7 +260,7 @@ syso_race_ethnicity_excl <- list(
 
 syso_veteran_pops <- list(
   "Veteran" = "Veteran",
-  "Non-Veteran" = "NonVeteran"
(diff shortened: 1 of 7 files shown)


hardcodes.R
```diff
@@ -260,7 +260,7 @@ syso_race_ethnicity_excl <- list(
 
 syso_veteran_pops <- list(
   "Veteran" = "Veteran",
-  "Non-Veteran/Unknown" = "NonVeteran"
+  "Non-Veteran" = "NonVeteran"
```
alex-silverman, line 263: I thought we wanted to combine non veteran and unknown? That's why we had this label. If we change it here, we have to change it throughout

alex-silverman: If you prefer Non-Veteran/Unknown, then we can make that work. I was able to make it work in my misc-fixes branch:
https://github.com/abtassociates/eva/blob/misc-fixes/system_composition_server.R#L155

image

alex-silverman (changes requested):
only change is to bring back "Non-Veteran/Unknown" in the hardcodes.R. You can also update the system_composition_server.R script to reflect the "Unknown" (see get_selection_cats() function).

alex-silverman (approved):
Looks good to me!

PR #600, opened 2024-09-20
Sankey fixes

alex-silverman:
- move sankey bar numbers to the labels, outside the bars 9for better readability)
- narrow the width of the sankey chart


kiadso (commented):
just needs updated snapshots

PR #603, opened 2024-09-24
Change lookback

kiadso:
not dropping lookbacks if they don't align to the project type input filters -AND-
dropping lookback enrollments with ExitDates older than the eecr's EntryDate minus 2 years..


alex-silverman (approved):
looks good to me other than the testing conflicts. Have you looked into how to resolve those?

PR #607, opened 2024-09-25
Waterfall labels

kiadso:
fixed plot label placement


vlopez0603:
I really dig how the lines for Active at Start/End appear based on the thickness of those two bars! Super nice touch. 

A couple additional thoughts:
- Can we get a bit more space between the data labels for Inflow and Outflow and the tops of the bars?
- When either Housed or Homeless is 0 at Active at Start and/or Active at End, can the data label be suppressed? We would still want to display a data label of 0 for the Inflow and Outflow categories.
  - Context: This is easiest to see when "Non-Residential" is selected as the Project Type Group (image below). By design there will be no one counted as Housed at Active at Start or Active at End. Currently, the chart displays a 0 in this scenario, but it's not especially clear that this number is related to the "Housed" category. Suppressing the data label if it's 0 for Active at Start/End (regardless of what's selected in the Project Type Group filter) would make this less confusing.
image (8)

kiadso:
Hi @vlopez0603, pull again and see if these edits resolve your feedback! Thank you!

vlopez0603 (approved):
I made a slight change to the vjust. Data suppression of 0 values for Active at Start/End works great. Approved!

PR #608, opened 2024-09-25
fixing how overlaps are handled. Some enrollments were seen as overla…

alex-silverman:
…pping when they really weren't. See e.g. personal ID 581289 in ICF-good, enrollment 860759 being marked as overlapping 857478


kiadso (approved):
I made a couple of small edits that probably didn't affect anything, but let me know @alex-silverman

PR #609, opened 2024-09-26
update more fonts to use hardcoded values and update ppt export fonts.

alex-silverman:


system_inflow_outflow_server.R
```diff
@@ -245,22 +245,9 @@ get_system_inflow_outflow_plot <- function(id) {
       size = sys_chart_text_font,
       inherit.aes = FALSE
     ) +
-    geom_text(
```
vlopez0603, line 248: I think something about the deletion of this code makes it such that the data labels for the inflow and outflow categories (inner bars of the chart) no longer appear at the top of the bars. In fact, the data labels for inflow and outflow do not show at all. Staging currently has the correct appearance for the data labels for the waterfall chart with numbers at the top of the bars for Inflow and Outflow and numbers at the sides of the bars for Active at Start/End.

vlopez0603 (changes requested):
The numbers for Inflow and Outflow no longer display at the top of the bars in the waterfall chart.

system_inflow_outflow_server.R
alex-silverman: i cannot for the life of me figure out how/when this code got dropped. but yea, pretty sure this is key

alex-silverman: I don't really even understand how you're highlighting this. It's not in the set of changes...

alex-silverman: okay, so I think i may have deleted this in a conflict resolution when I tried merging with phase2-dev. Maybe I hadn't pulled the latest phase2-dev or i may have done the conflict resolution wrong. Either way, adding it back in.

vlopez0603 (approved):

PR #614, opened 2024-09-30
adding back the metadata tab that somehow got removed from System Dem…

alex-silverman:
…ographics


vlopez0603 (approved):
Confirmed the data export for the System Demographic chart included the metadata tab. Updated the name of the tab itself to be in line with the chart's name and to match the naming convention of the metadata tabs in data exports for the waterfall and sankey chart.

PR #616, opened 2024-09-30
Unspecified 3 yrs

kiadso:
overwrites inflow/outflow categories if the data was run on less than three years (or 1094 days) in the waterfall and the inflow/outflow download


alex-silverman:
It looks like the tests are passing because the structure of those tests has changed. They're all being stored in the 4.4 folder for some reason, rather than a subfolder for each test (e.g. main-valid).

alex-silverman (changes requested):
I think we should do some code consolidation here

PR #619, opened 2024-10-01
Phase 2 Metadata Capture for System Overview

zabalad:
The following metadata will pull into the current metadata report:

- High Priority. Usage statistics: weekly and monthly primary and secondary horizontal tab hits
- High Priority. Weekly and monthly System Overview downloads 
  - High Priority. By total
  - Lower Priority. By chart (more like a nice-to-have if things get gnarly)
    - Break out by download type (image vs tabular)
- High Priority. Demo mode stats (primary and secondary horizontal tabs visited last 6 months, weekly downloads by chart)
- Low Priority. Count of inclusive versus exclusive methodology filter changes

All high priority items have been committed. Lower priority items will be added if time allows.

### DIFF
--- a/helper_functions.R
+++ b/helper_functions.R
@@ -257,7 +257,8 @@ logSessionData <- function() {
     SourceContactFirst = Export()$SourceContactFirst,
     SourceContactLast = Export()$SourceContactLast,
     SourceContactEmail = Export()$SourceContactEmail,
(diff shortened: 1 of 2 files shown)


kiadso (approved):
tested with back to back uploads that everything looks good in the app

PR #620, opened 2024-10-02
Phase 2 metadata capture - clarifying info

zabalad:
Clarified which primary tab corresponds with the subtab being clicked. Also added primary tab name to the tabular data download and added title slide to the image (ppt) download.


alex-silverman (approved):
Nicely done. By referencing input$syso_tabbox, this makes things more dynamic, and by putting it in the system_overview_server, it makes things more concise

PR #627, opened 2024-10-07
speed up the processing after the file is done processing. The extra …

alex-silverman:
…time that everything takes to load seems to be directly resulting from the observeEvent that creates and stores the inflow_outflow_df.


zabalad:
Compared to dev branch (uploaded good.zip in both branches). Can confirm the app isn't breaking. Some differences noted:
- In the speedups branch I see "Total Homeless," which I didn't see in dev. Victoria confirmed Total Homeless should be removed.
image

-  the dev system demographics report includes "Unknown" and the values for "Total" are different, whereas the speedups branch doesn't include "Unknown" and has a lesser value for "Total." Victoria confirmed Unknown should be included.

alex-silverman:
> Compared to dev branch (uploaded good.zip in both branches). Can confirm the app isn't breaking. Some differences noted:
> 
> * In the speedups branch I see "Total Homeless," which I didn't see in dev. Victoria confirmed Total Homeless should be removed.
> * the dev system demographics report includes "Unknown" and the values for "Total" are different, whereas the speedups branch doesn't include "Unknown" and has a lesser value for "Total." Victoria confirmed Unknown should be included.

I just merged in dev. Can you confirm those things are updated?

zabalad:
> > Compared to dev branch (uploaded good.zip in both branches). Can confirm the app isn't breaking. Some differences noted:
> > 
> > * In the speedups branch I see "Total Homeless," which I didn't see in dev. Victoria confirmed Total Homeless should be removed.
> > * the dev system demographics report includes "Unknown" and the values for "Total" are different, whereas the speedups branch doesn't include "Unknown" and has a lesser value for "Total." Victoria confirmed Unknown should be included.
> 
> I just merged in dev. Can you confirm those things are updated?

Looks good! Can confirm "Total Homeless" is removed from System Flow and "Unknown" is added to System Demographics.

zabalad (approved):

PR #630, opened 2024-10-08
if HH type filter is AO or AC, make sure to include people that are A…

alex-silverman:
…O/UY or AC/PY


vlopez0603 (approved):
Change looks good. Thanks!

PR #643, opened 2024-10-28
Dev hotfix

alex-silverman:
- Handle # symbol in csv files
- correctly default to All genders by referencing choice value, not name


vlopez0603:
Confirmed that Client.csv and Export.csv no longer save in the Files panel after a failed upload. There are still a couple of noticeable things, though.

- Within the same session, going from successful to unsuccessful upload, you still see the org names from the previous upload in the dropdown menus on the View Client Counts and Org-level DQ pages. 
- Within the same session, going from successful to unsuccessful upload, on the System Performance Overview page, the Data and Image download buttons are visible but result in failed downloads. Ideally, those buttons would not be visible/clickable in this kind of scenario.

vlopez0603 (approved):
Approved. The latest push addresses the two issues in my previous comment:

- In testing, VL confirmed that when an unsuccessful upload follows a successful upload in the same session, the View Client Counts and Organization-level DQ pages no longer show org names from the previous successful download in the dropdown menus.
- In testing, VL confirmed that when an unsuccessful upload follows a successful upload in the same session, the System Performance Overview page now hides the Data and Image download buttons.

PR #652, opened 2024-11-13
522 zero utilization

kiadso:
this aims to resolve Issue 522. I know creating test datasets is not firmed up right now, but I wanted to go ahead and get this PR going


tests/update_test_good_dq.R
```diff
@@ -98,4 +98,23 @@ original_data_fixed_cols$Inventory <- original_data_fixed_cols$Inventory %>%
...
+  filter(EnrollmentID != "696923")
+
+original_data_fixed_cols$HealthAndDV <- original_data_fixed_cols$HealthAndDV %>% 
+  filter(EnrollmentID != "696923")
+
+original_data_fixed_cols$IncomeBenefits <- original_data_fixed_cols$IncomeBenefits %>% 
+  filter(EnrollmentID != "696923")
+
+original_data_fixed_cols$YouthEducationStatus <- original_data_fixed_cols$YouthEducationStatus %>% 
+  filter(EnrollmentID != "696923")
```
alex-silverman, line 116: Can we simplify this with a loop so it's clearer what we're trying to do? e.g.:
```
original_data_fixed_cols[tables_to_filter] <- lapply(
  original_data_fixed_cols[tables_to_filter],
  function(df) df %>% filter(EnrollmentID != "696923")
)
```

alex-silverman:
Results from running the tests:

1. There are 2 new PDDE warnings for zero utilization, which is expected.
2. There's one fewer client (enrollment 696923) appearing in the Client Count download, which is NOT expected. Still trying to figure out why this is happening.

alex-silverman (approved):
made a small change but this looks good!

PR #657, opened 2024-11-22
change the HP DQ Org Issues tab name to just say Issues because there…

alex-silverman:
… are fewer than 10 HP errors


kiadso (approved):
replicated your work, @alex-silverman onto the System DQ page as well- thank you!

PR #658, opened 2024-11-22
Feat/permit utf8

alex-silverman:


01_get_Export.R
```diff
@@ -9,6 +9,7 @@ list_of_problems <- list()
 # for each file in the csv, loop through the file names in the csv
 for (file in unique(cols_and_data_types$File)) {
   #import the csv and save it as a data frame
+  print(paste0("importing ", file))
```
alex-silverman, line 12: @kiadso - this is just so we have better logging of steps. may help us debug issues more quickly

.gitignore
```diff
@@ -1,4 +1,4 @@
-*.Rproj.user
+.Rproj.user
```
alex-silverman, line 1: @kiadso - R studio seems to want .Rproj.user. It added it automatically.

fsa_server.R
```diff
@@ -0,0 +1,120 @@
...
+  for (col in names(dt)) {
+    if (is.character(dt[[col]])) {
+      # Detect non-UTF-8 characters
+      invalid_utf8_rows <- which(!stringi::stri_enc_isutf8(dt[[col]]))
+      if (length(invalid_utf8_rows) > 0) {
+        problematic_characters$utf8_issues <- lapply(invalid_utf8_rows, function(row) {
+          list(
+            row = row, 
+            col = col,
+            chars= iconv(
```
alex-silverman, line 25: This converts windows-encoded characters, like smart-quotes, to the UTF-8 approximation so that the user will see the actual character in the export. Otherwise, they'd see the UTF-8 interpreted.

NOTE 12/13/24: AS and GD decided better to silently convert and not flag because the actual characters may be legitimate UTF-8 characters (like smart quotes are), but the FILE may be encoded as Windows. Helping the user encode a file as UTF-8 will be difficult and it's not their fault in the first place since they just want the characters that ARE VALID UTF-8 anyway

kiadso (changes requested):
A couple of things I found in my testing/looking around:

1. In a different sample dataset I created on my own, Eva is missing the non-UTF character I added to Project.csv$ProjectName. No flag is thrown on it.
2. Further, it sees the character in its non-UTF format in some cases, and changes the character to a UTF translation in other contexts, causing unexpected behavior in the Client Counts report, and possibly other calculations. 
3. I think the current guidance text as it is: "See page 4 of the HMIS CSV Specs for characters permitted in string fields. Contact your vendor to ensure impermissible characters are excluded from your export." needs to be changed since Eva is no longer going strictly by the FY2024 HMIS CSV Export Specifications in anticipation of the FY2026 changes and now allows the UTF-8 character set. I'm thinking we should change this guidance to: "The HMIS CSV Specs restrict the use of some characters in string fields. To locate the character(s) that are flagging this issue, please run and download the Impermissible Character Detail. Contact your vendor to ensure impermissible characters are excluded from your export."
4. I do notice that the sample dataset that you're using does trip the flag and produces the workbook output as expected, but once you've uploaded the file as it is, if you go to Organization-level DQ and select Organization [B and click "Top Ten Projects" on General Errors, you will see an error there: "invalid UTF-8 byte sequence detected; try calling stri_enc_toutf8()" so Eva isn't gracefully handling the bracket or other character(s) somewhere.
I'll send along the sample dataset I'm using so you have it! Thank you @alex-silverman!

alex-silverman:
> A couple of things I found in my testing/looking around:
> 
> 1. In a different sample dataset I created on my own, Eva is missing the non-UTF character I added to Project.csv$ProjectName. No flag is thrown on it.
> 2. Further, it sees the character in its non-UTF format in some cases, and changes the character to a UTF translation in other contexts, causing unexpected behavior in the Client Counts report, and possibly other calculations.

I figured out the problem and have pushed a fix. The short of it is that your Project.csv file was already UTF-8 encoded, but my convert-Windows-to-UTF8-so-characters-display-nicely code still ran and messed things up. The fix is to only _conditionally_ do that conversion if the file isn't already UTF-8.

I think this stuff is fascinating for some weird reason, so below is a very detailed explanation of what I found. **_FEEL FREE TO SKIP!...._**

There are a couple things going on here. The character you added, ‰, is a valid UTF-8 character, just like smart quotes are. However, the encoding of the files is the key. *My* test zip ("S:\projects\CE_Data_Toolkit\Data Sets\FY24-ICF-good_non_utf8.zip") - which I've updated to have one ProjectName be the same as yours for the sake for clarity - is encoded as WINDOWS-1252/ANSI/ISO-8859-1/Latin-1. (If you open the Project.csv file in Notepad and look at the bottom-right corner you'll see it's encoded as ANSI). *Your* test zip ("S:\projects\CE_Data_Toolkit\Data Sets\FY24-ICF-hashed-bad-non-utf.zip") is UTf-8.

When R reads *my* _WINDOWS-1252_ _encoded_ file, it sees the _Windows_ byte representation of that character, `\x89`. If you were to print the ProjectName to the console, you'd see: `Organization M - TH\x89`. Since `\x89` is not a valid UTF-8 byte, it doesn't get automatically converted like all the other regular ASCII characters (a, b, c, etc.).

On the other hand, when R reads *your* _UTF-8_ _encoded_ file, it sees the _UTF-8_ byte representation, `0xE2 0x80 0xB0`. If you were to open it in Excel, **which forces a Windows-1252 interpretation**, the bytes are interpreted as 3 separate Windows-encoded characters:

  0xE2 → â
  0x80 → €
  0xB0 → °

so you see `Organization M - THâ€°`. But because R assumes UTF-8, the 3 bytes correctly get interpreted as ‰, so if you print to console, you see `Organization M - TH‰`.

The problem comes when my convert-Windows-to-UTF8-so-characters-display-nicely code - which runs at the end of processing - tries to display the actual characters, not the weird encoded versions of them. Here's the key code: `iconv("string-to-convert", from = "WINDOWS-1252", to = "UTF-8")`.

Before I get into how it was wrong, here's how it works, as written: it takes the byte representation of "string-to-convert", interprets it as WINDOWS-1252, and then converts the resulting characters to UTF-8. For the ASCII characters, the byte interpretations are the same, so the characters are the same and the code does nothing. Beyond those, things happen. Here's how it works with the Windows-encoded file: it sees `\x89`, interprets it as WINDOWS-1252, which is ‰, then converts that character to UTF-8, which is still ‰. But with *your* file it sees ‰ as in the default UTF-8 encoding `0xE2 0x80 0xB0`, interprets as Windows-1252 (i.e. â€°) and then converts those _characters_ to their UTF-8 equivalents, which are just themselves. That's why you see those characters in the dropdown!

> 3. I think the current guidance text as it is: "See page 4 of the HMIS CSV Specs for characters permitted in string fields. Contact your vendor to ensure impermissible characters are excluded from your export." needs to be changed since Eva is no longer going strictly by the FY2024 HMIS CSV Export Specifications in anticipation of the FY2026 changes and now allows the UTF-8 character set. I'm thinking we should change this guidance to: "The HMIS CSV Specs restrict the use of some characters in string fields. To locate the character(s) that are flagging this issue, please run and download the Impermissible Character Detail. Contact your vendor to ensure impermissible characters are excluded from your export."

I've made this change in EvaChecks.csv

> 4. I do notice that the sample dataset that you're using does trip the flag and produces the workbook output as expected, but once you've uploaded the file as it is, if you go to Organization-level DQ and select Organization [B and click "Top Ten Projects" on General Errors, you will see an error there: "invalid UTF-8 byte sequence detected; try calling stri_enc_toutf8()" so Eva isn't gracefully handling the bracket or other character(s) somewhere.

This was due to when my conversion code ran, which was after processing. But by this time, prep datasets have been created with the "bad" characters. The fix is to simply due the conversion code asap, so I added it to the importFile function.

>    I'll send along the sample dataset I'm using so you have it! Thank you @alex-silverman!

01_get_Export.R
kiadso: I really like this! thank you!

kiadso (approved):
Hi- thank you so much for all your work on this @alex-silverman. All my testing conditions I named in my last review came back as resolved and it all appears to work great. I do have the one .gitignore question but it is super minor and we can either discuss and push to here before merging or discuss and push to another branch if it's something we want. Thanks again!

PR #666, opened 2024-12-18
Feat/client level export

alex-silverman:


kiadso (commented):
I know this is in draft, but I had this PR as one to review and drop comments if I have them, and I do have a couple of them-

1. I'm aware there are edits coming to some of the column names
2. I'm aware also that there are edits coming (maybe) to displaying ExitDate or ExitAdjust, so however that lands is fine either way
3. I did notice that ProjectType values are coming through as numeric, but if you're so inclined, you can use the project_type_abb() function which will translate those numbers to easier to understand common abbreviations for those project types.
4. In the metadata worksheet, the date formats are inconsistent. In general Eva displays either yyyy-mm-dd or mm-dd-yyyy formats but never mm-dd-yy, so I think making them all either yyyy-mm-dd or mm-dd-yyyy would be good.
5. This is so tiny, but in general since we read left to right and view things on the left as "past" and things on the right "next", placement of LivingSituation (which is really Prior Living Situation) and Destination should be reversed since LivingSituation is an earlier situation and Destination is the later situation.
6. I notice DV Category is listed but we're not displaying that in Eva proper- just flagging this because the reasoning behind not including it was those categories were not vetted by HUD's lived experience resources.

Everything looks really great and this will be a fantastic addition to Eva! Thanks @alyssandrichik

PR #668, opened 2024-12-19
only flag missed movein if HoH

alex-silverman:


kiadso (approved):
this needs a changelog entry but everything else looks good.

PR #672, opened 2024-12-31
PDDE check updates

zabalad:
Fixed problems with PDDE checks caught by shinytest


alex-silverman (approved):
In an effort to get this in by EOW, I updated the snapshots to handle the conflicts with Dev. I also added a couple other process-y things that have come up a few times: ignoring lind-endings in case we switch from Windows to RStudio/Linux and ignoring .Rproj.

PR #687, opened 2025-01-19
Overlap fixes

alex-silverman:
This PR fixes sveral things:
1. fixes warning about DateProvided when Services file is empty and simplify some code around variable selection using existing variable vectors
2. fix edge case in NbN vs. NbN overlap detection - if too many dups, the join results in more rows than expected and it crashes eva.


04_initial_data_prep.R
```diff
@@ -102,7 +102,39 @@ EnrollmentStaging <- Enrollment %>%
...
+    min_AgeAtEntry = min(AgeAtEntry)
+  ) %>%
+  mutate(
+    HouseholdType = factor(
+      ifelse(
+        any(between(AgeAtEntry, 0, 17)) & max_AgeAtEntry >= 18,
+        ifelse(
+          between(max_AgeAtEntry, 0, 24),
+          "PY",
+          "AC"
```
kiadso, line 118: Since PY is a subcategory of AC, I feel like this should be named something like ACminusPY or some other way of being sure that we're not saying like if x is true, then "orange", otherwise "fruit". Like we should call it "orange" and "fruitsexceptoranges" as an example. Same goes for "UY" being a subcategory of "AO". It should be "UY" and "AOminusUY" or something similar.

kiadso (commented):
I have to flip to SSVF for now, and when I come back to this I'll finish reviewing, for now, just submitting this as a comment

04_initial_data_prep.R
alex-silverman: I totally hear you. We had something like that before. I don't feel super strongly. On the one hand, I like that it currently aligns with the choices the user actually selects from in the System Overview filter, which is mutually exclusive, as well as how we report on them in the DQ Export Overlap Details. On the other hand...you're right.

Rminor_elevated.Rproj
alex-silverman, line 1: try removing this from the PR

04_initial_data_prep.R
alex-silverman: After talking with Diana, we decided to go with Genelle's suggestion. At the very least, this makes the line where we collapse HouseholdType down for the purposes of the DQ Export, more intuitive:

c("PY","AC") ~ "AC"   vs.   c("PY","ACminusPY") ~ "AC"

The latter is a lot less confusing.

PR #692, opened 2025-01-22
Speedups

alex-silverman:


tests/update_test_good_fsa.R
```diff
@@ -3,8 +3,11 @@ print("updating test good for FSA")
 # Incorrect Date Format, High-Priority (check 11)
 reduced_data_fsa$Client$DOB <- format(reduced_data_fsa$Client$DOB, "%d-%m-%y")
 
-# Incorrect Date Format, Error (check 47)
-reduced_data_fsa$Client$DateUpdated <- format(reduced_data_fsa$Client$DateUpdated, "%d-%m-%y")
+# Incorrect Datetime Format, Error (check 47)
+reduced_data_fsa$Client$DateUpdated <- "31-03-21"
+
+# Incorrect non-date format
```
alex-silverman, line 9: change this comment to indicate non-*numeric*

alex-silverman (commented):
Need to confirm that the test-fsa-001.json change from HP to Error is correct. Also pay attention to the text of the DOB data type check. Seems off. Same with the AssessmentType check

alex-silverman:
> Need to confirm that the test-fsa-001.json change from HP to Error is correct. Also pay attention to the text of the DOB data type check. Seems off. Same with the AssessmentType check

this was actually a bugfix. The bug happened when constructing the FSA sub-datasets that highlight unexpected data types, we were starting with problems - which tells us which columns had unexpected types and their actual column numbers in the dataset - and merging with columns.csv, which tells us the expected columns' number and data type. We use the latter to grab the column name. If the file was missing a column, like when we remove BlackAfAmerican (an HP col) and DateCreated (a non-HP col) from the Client.csv file in update-test-good-fsa.R, then, e.g., DischargeStatus goes from being the 41st column to the 40th column and DateUpdated goes from 43rd to 41st. The `problems`  dataset looks like this:

| row | col | expected       | actual   | file                                                                 |
|-----|-----|----------------|----------|-----------------------------------------------------------------------|
| 2   | 9   | date in ISO8601 | 19-01-97 | /media/sdrive/projects/CE_Data_Toolkit/Eva_AlexS/eva/Client.csv       |
| 2   | 41  | date in ISO8601 | 03-11-21 | /media/sdrive/projects/CE_Data_Toolkit/Eva_AlexS/eva/Client.csv       |

Here, 41 is the DateUpdated column, which, in update-test-good-fsa, we intentionally reformatted to %d-%m-%y, which is not the expected format. So it's correctly flagging this problem. However, when we merge with columns.csv, 41 is still the DischargeStatus column, so we end up reporting that "Please check that the DischargeStatus column in the Client file has the correct date format.", which is wrong/confusing.

The new code, however, starts with the columns.csv to get the column names and expected data types. Then it looks up the actual data types from the dataset.

PR #694, opened 2025-01-23
hotfix - changelog

zabalad:
Revised changelog to reflect update to inventory pdde check


alex-silverman (changes requested):
the csv files in tests/helper_data/main-valid should, in theory, never be .new. That means that something changed about the data in those csvs. Of course, in this case, the changelog changes obviously didn't cause that, but we don't want to start tracking those files either. So, at the very least, we should delete those files. But I am curious why they're seen as new.

alex-silverman (approved):

Issue #680, opened 2025-01-08
Bed Type incompatible with Housing Type Warning Appearing for Programs with Consistent Bed Types and Housing Types


entomacy:
Bed Type incompatible with Housing Type Warning Appearing for Programs with Consistent Bed Types and Housing Types

**Time of Issue**
1/7/25 9:30am ish

**Expected behavior**
We are seeing programs with this warning in the PDDE report that we would not expect to see. Guidance says, "ES projects must record a Bed Type that is logically consistent with Housing Type at the project level. Facility-based beds align to the Housing Type of site-based and voucher-based beds align to the Housing Type of tenant-based." Our sample program is **Portage - Family and Community Services - Housing & Emergency Support Services - ES Overflow**. This program is site-based with facility-based current inventory records. We suspect that the warning is coming from inventory records that are ended and not in the reporting range (2013-2020 inventory records.)

vlopez0603:
Hi @entomacy - This should have been resolved in our most recent update. Please submit a new ticket if you are still experiencing the same issue.

PR #699, opened 2025-01-30
Reference active inventory for pdde check

zabalad:
Reference active inventory records instead of all inventory records to check bed type compatibility with housing type. Addresses issue 680: https://github.com/abtassociates/eva/issues/680


alex-silverman (approved):
looks good!

PR #702, opened 2025-02-04
Reject export to prevent crash

zabalad:
- Currently, if Export.csv has more than 1 row, it will crash Eva
- Add FSA check to the 00_invalid script to reject the file if it has more than 1 row.
- ----Otherwise, make it HP so it gets rejected
- ----Also, will need the other parts of an issue (Guidance, Type (i.e. Error or High Priority)). Reminder, these get added to the eva_checks.csv file (the .xlsx version is deprecated)
- Also, add an automated test to check that if the Export.csv has > 1 row, it will successfully be flagged

ACCEPTANCE CRITERIA

- An Export.csv with more than 1 row should be handled gracefully/correctly. It shouldn't crash the site.


00_initially_valid_import.R
```diff
@@ -116,6 +116,12 @@ if(tolower(tools::file_ext(upload_filepath)) != "zip") {
       title = "Unsuccessful Upload: You uploaded an unhashed data set"
     )
     logMetadata("Unsuccessful upload - not hashed")
+  } else if(importFile(upload_filepath, "Export") %>% nrow() != 1) {
```
alex-silverman, line 119: Let's use the already-created `Export()` reactiveVal, instead of re-importing the file. And since we're telling them the problem is that the file has more than 1 row, let's change the check to be >1, instead of !=1. Otherwise, if it has 0 rows, the pop-up wouldn't be right.

00_initially_valid_import.R
```diff
@@ -116,6 +116,12 @@ if(tolower(tools::file_ext(upload_filepath)) != "zip") {
...
+    show_invalid_popup(
+      issueID = 140,
+      title = "Unsuccessful Upload: You uploaded an Export.csv file containing more than 1 row. Export.csv should only have one row."
```
alex-silverman, line 122: For consistency, let's change the "should only have one row" to "should only have 1 row"

alex-silverman (changes requested):
a couple super small changes. Let's also update the changelog

alex-silverman:
Let's also update the changelog. I'll consolidate all the changelog entries for the coming push once all the PRs are in. But it would be helpful to have the changelog text in this PR, even if the date is not quite right.

alex-silverman (approved):

PR #712, opened 2025-02-11
use ReportStart() in the call to homeless_cls_finder within active_at…

alex-silverman:
…_Start_homeless definition


zabalad (approved):
Looks good

PR #717, opened 2025-02-18
Memory cleanup

alex-silverman:
remove uploaded zip file from Rtmp directory


zabalad (approved):
Everything works as expected. Once tests pass, this should be good to merge.

PR #718, opened 2025-02-18
move the logMetadata call that tracks System Overview-related Data Downloads...

alex-silverman:
... from the sys_export_summary_initial_df - which is called in a number of places, including the PPT downloads, and therefore we're overcounting tabular downloads - to within the download handler of the specific Data Download. This is more intuitive and doesn't overcount downloads

### DIFF
--- a/system_composition_server.R
+++ b/system_composition_server.R
@@ -664,6 +664,8 @@ output$sys_comp_download_btn <- downloadHandler(
       col_names = TRUE
     )
     
+    logMetadata(paste0("Downloaded System Overview Tabular Data: ", input$syso_tabbox,
+                       if_else(isTruthy(input$in_demo_mode), " - DEMO MODE", "")))
     exportTestValues(sys_comp_report_num_df = num_df)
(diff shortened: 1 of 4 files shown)


zabalad (commented):
@alex-silverman - if code to capture download of system overview tabular data is in 3 different places, doesn't that mean we would log a single download 3 consecutive times? Maybe I'm misunderstanding how this works

alex-silverman:
> @alex-silverman - if code to capture download of system overview tabular data is in 3 different places, doesn't that mean we would log a single download 3 consecutive times? Maybe I'm misunderstanding how this works

which 3 places are you referring to? the key issue was that we were logging a Downloaded Tabular Data message within the `sys_export_summary_initial_df` function, which got called not only during PPT downloads, but also Tabular downloads and even the Client Export download. Do a CTRL+SHIFT+F search for that function name in Main and you'll see all the places we called it. On top of that, we were also logging the Sys Inflow Outflow tabular download explicitly, but not doing that with the other 2 charts.

alex-silverman:
merging this branch into the metadata-early-birds branch

Issue #724, opened 2025-02-20
VSP projects incorrectly tagged as containing client data


jessefjorstad:
**DO NOT UPLOAD YOUR EXPORT TO GITHUB (HASHED OR NOT)**

**Describe the bug**
I uploaded 4 different CSVs, one per project type (ES, TH, RRH and PSH). And downloaded the PDDE reports. In reviewing them, there were projects that were identified as being VSP (accurate) and containing client data (incorrect). I filtered the Enrollment CSVs by the implicated Project IDs (respective to the project types), but the Project IDs were not found in the Enrollment CSVs. This occurred with the ES, TH and RRH files.

**Time of Issue**
Yesterday (2/20/2025) morning

**To Reproduce**
Unknown. Standard upload of CSV

**Expected behavior**
That no projects that have no enrollments would be identified as having enrollments

**Screenshots**
Note: I consolidated all of the PDDE downloads into one sheet:

![Image](https://github.com/user-attachments/assets/72c6b686-6820-451d-a21d-93ffce1df710)

From the TH enrollment file. You can see that 3735 is not listed

![Image](https://github.com/user-attachments/assets/98bc1550-2042-4129-8506-6528a091054a)

**Desktop (please complete the following information):**
 - Operating System: Windows 10
 - Browser: Chrome

**Smartphone (please complete the following information):**
N/A

**Additional context**
**DO NOT UPLOAD YOUR EXPORT TO GITHUB (HASHED OR NOT)**

vlopez0603:
Hey @jessefjorstad - We're currently investigating this issue. We'll let you know when we anticipate it being resolved.

PR #727, opened 2025-02-28
Hotfix/vsp

alex-silverman:
Mainly handles #724 

However, 3 additional, back-end-y edits:
- Accounting for a side-effect of more recent versions of data.table::fread, which types Date variables as IDate. The fix, implemented here, is to automatically convert them back to Dates. An alternative, slightly better solution, is to update the code to look for IDate wherever it currently looks for Date.
- updates helper data files with the removal of the DomesticViolenceCategory (since we aren't using it), which should have happened for a previous PR but I forgot.
- deleted snapshot pngs. we're gitignoring them now because theyre not necessary to carry through to github. This should save some space, too.


alex-silverman:
@kiadso - tagging/assigning to you also, in case you have time, since this is a super simple fix. But no worries if you don't!

PR #734, opened 2025-03-14
Display adjustments per HUD's guidance

alex-silverman:


zabalad (changes requested):
Looks good to me! Waiting for confirmation of review and edits to columns.csv from Alex

PR #740, opened 2025-03-20
Trying out a very obvious, but minor change to see if github tests wi…

vlopez0603:
…ll run and pass in PR.


alex-silverman:
In case it saves any more time working on this- I'd been dealing with this failure right before I left. I was messaging Theo about it on the R Studio teams channel if you want more info. It seems Chromote, which shinytest2 uses to test the app, was recently upgraded in a way that now fails to open Chrome. I found a fix that can be adapted for the GithubTests, but didn't have the chance to implement it yet.  Can work on that next week.

PR #741, opened 2025-03-21
Allow bed type "Other" for site based ES projects

zabalad:
For ES projects, if HousingType is 1 or 2, then we should allow for ESBedType 1 and 3


vlopez0603 (approved):
Looks good!

alex-silverman (approved):
programmatically, looks good to me! Let's just get the tests passing

Issue #738, opened 2025-03-18
System Performance Overview - "No date to show"


TinoPazHSN:
**Describe the bug**
Uploaded CSV Export - from 3/1/2024 to 2/28/2025

**Time of Issue**
03/18/2025, Eastern

**To Reproduce**
1. Upload CSV Export
2. Review - **Assess Data Quality** view: everything OK
3. Review **System Performance Overview** - receive message saying "No data to show"

**Expected behavior**
System Performance Overview visualizations.

**Screenshots**

![Image](https://github.com/user-attachments/assets/042d8bfd-2466-47f7-8cb4-13fcac2a82c0)

**Desktop (please complete the following information):**
 - Operating System: Windows 11 Pro
 - Browser: Microsoft Edge (Version 134.0.3124.72)

vlopez0603:
Hi @TinoPazHSN - Our team will investigate this and reach out to you soon.

desireeblakeabt:
Hi @TinoPazHSN and thank you for the question. The visualizations are suppressed when there is not enough data to populate. Specifically, the System Performance Overview will display "no data to show" when there are less than 11 clients. I can see that your reporting period is over a year so please verify the following before we investigate further. Was the report pulled on a single program or for more than one program? Is there data for more than 11 clients populating for this report? You can easily verify this by selecting 'client level download' on the System Performance Overview page and viewing the row for 'Total Served (Start + Inflow) People'.
 
If this is showing less than 11 clients, then Eva is functioning correctly and will require more data to generate the visualizations. If there is more than 11 clients, please let us know so we can investigate as this may highlight a larger error.

TinoPazHSN:
Thank you for following-up on this issue.  Actually, it was a CoC wide data pull for a whole year!


[https://ci3.googleusercontent.com/mail-sig/AIorK4yT-9nzNptSTaHmpxwhiPe9XNCd9xtDA5soM0PDRQIjhcGVODRTuv4XQOGWI6dNz34cv80aRSU]

Agustin "Tino" Paz ǀ HMIS Senior Data Analyst

Ph:  (407) 893 - 0133 x650

Fax: (407) 893 - 5299

142 East Jackson Street

Orlando, FL 32801

www.hsncfl.org<http://www.hsncfl.org/> | www.cfch.org<http://www.cfch.org/>

Have an HMIS Question?
Get a guaranteed response within 48 business hours!
Send your question to ***@***.***

[...]

PR #742, opened 2025-03-26
use lubridate::dyears instead of lubridate::years because the former …

alex-silverman:
…counts years in days, rather than just subtracting the number from the year value, which fails for leap years if the number is not 4, since, e.g. 2 years prior to a leap year doesn't contain a 02-29 date.


alex-silverman:
Fixes #738

zabalad (approved):
Looks good! I didn't notice any issues when I ran the app. Approved.

PR #747, opened 2025-03-26
- better memory handling by closing session after a timeout and doing…

alex-silverman:
… gc(). However, because when the session ends, the disocnnect pop-up is displayed saying Eva has crashed, we handle that explicitly by hiding the pop-up.

- for better organization, moving the JS code into a separate script


zabalad (approved):
This looks good to me, approved!

PR #763, opened 2025-04-22
Hotfix/pdde check

alex-silverman:
account for scenario where there's no available beds in Active Inventory w No Enrollments check


mlfurman3 (approved):
Logic looks good to me, just using a function to return both cases instead of an if statement.

PR #764, opened 2025-04-22
Fix/move in date adjust

alex-silverman:


alex-silverman:
Here's a summary of the latest version of the code:

- First we calculate an adjusted Household Move-In Date (HMID). This is based solely on the HoH's MoveInDate. The adjustments follow similar rules as before:

1.	If 
a.	PSH/OPH enrollment AND
b.	Entry Date < 10/1/2017 AND 
c.	Move-In Date = NULL or Move-In Date < Entry Date 
**Set to EntryDate**

2.	If
a.	PSH/OPH enrollment with AND
b.	Entry Date < 10/1/2017 AND 
c.	Move-In Date >= EntryDate 
**Set to MoveInDate**

3.	If 
a.	Non-PH project type OR
b.	MoveInDate is invalid (i.e. before EntryDate or after (or equal to if PSH/OPH) ExitAdjust)
**Set to NA**

If there are (erroneously) multiple HoH, take the earliest one.

- Next, we use HMID to set the value of MoveInDateAdjust for all household members. 

1. For HoH, it's simply the HMID
2. For non-HoH:
a. !ProjectType %in% ph_project_types, NA_Date_,
b. EntryDate > HMID, EntryDate,
c.  EntryDate <= HMID, HMID,
d.  ExitAdjust < HMID, NA_Date_

04_initial_data_prep.R
```diff
@@ -356,8 +317,6 @@ EnrollmentAdjust <- Enrollment %>%
       )
   )
 
-rm(HHMoveIn)
-
```
mlfurman3, line 360: Do we no longer need to remove HHMoveIn?

mlfurman3:
I think this logic makes sense to me, made one minor comment and otherwise will just need the conflicts to be resolved.

04_initial_data_prep.R
alex-silverman: @mlfurman3 i think we should still remove it, since it's only used as a temporary dataset here.

alex-silverman:
tests passed so I'm merging

PR #779, opened 2025-05-22
776 quoted csvs are being treated as strings for currentlivingsituationcsv

alex-silverman:


alex-silverman:
@vlopez0603 - can you review updated EvaChecks guidance text?

public-resources/EvaChecks.csv
vlopez0603, line 1: As hysterical as I think "Beware!" is--like these hidden characters are just laying in wait until the perfect moment to strike--I suggest we make the slight change to, "Hidden, zero-width space characters may cause Eva to interpret numeric variables as character variables."

vlopez0603 (changes requested):
Suggesting a slight change to the guidance language to, "Hidden, zero-width space characters may cause Eva to interpret numeric variables as character variables."

public-resources/EvaChecks.csv
alex-silverman: prefixed with "Note that...", so it's clearer that it's only one to look out for.

PR #785, opened 2025-06-26
Metadata downloads

mlfurman3:
Moved "Download Client Level Export" event from DQ to Sys Org in metadata analysis. Also changed the read-in code from dplyr to data.table so it is much faster to run the Rmd.

Note: This shifted the color palettes since they were assigned to events alphabetically. I updated the color palettes used in the charts to match the Eva branding guide while moving some lighter colors to the end of the list so only used as necessary.

### DIFF
--- a/changelog.R
+++ b/changelog.R
@@ -2,7 +2,7 @@ output$changelog <- renderDT({
   tribble(
     ~ Date,
     ~ Change,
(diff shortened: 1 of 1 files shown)


alex-silverman (approved):
this looks good to me! Let's merge!

PR #786, opened 2025-06-26
Bslib mbm

mlfurman3:
has some month-by-month changes but is outdated - will merge mbm into this branch before review.

### DIFF
--- a/system_inflow_outflow_server.R
+++ b/system_inflow_outflow_server.R
@@ -901,8 +901,8 @@ get_sys_inflow_outflow_monthly_plot <- function(isExport = FALSE) {
     averages <- fmean(plot_data$Count, g=plot_data$PlotFillGroups)
     
     # Inflow and Outflow here include Inflow+Active at Start and Outflow+Active at End
-    totals_start <- fsum(plot_data[InflowOutflow == "Inflow", Count])
-    totals_end <- fsum(plot_data[InflowOutflow == "Outflow", Count])
+    totals_start <- fsum(plot_data[PlotFillGroups == "Inflow", Count])
+    totals_end <- fsum(plot_data[PlotFillGroups == "Outflow", Count])
     avg_monthly_change <- (totals_start - totals_end)/(length(session$userData$report_dates) - 1)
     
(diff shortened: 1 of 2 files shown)


tests/testing_functions.R
```diff
@@ -489,11 +548,11 @@ main_test_script <- function(test_script_name, test_dataset) {
     customDownload(app, "downloadSystemDQReport", "System-DQ-Download.xlsx")
     customDownload(app, "downloadOrgDQReport", "Org-DQ-Download.xlsx")
     customDownload(app, "sys_inflow_outflow_download_btn", "System-Flow-Download.xlsx")
-    customDownload(app, "sys_inflow_outflow_download_btn_ppt", "System-Flow-Download-PPT.pptx")
+    #customDownload(app, "sys_inflow_outflow_download_btn_ppt", "System-Flow-Download-PPT.pptx")
```
alex-silverman, line 551: why are these commented out?

global.R
```diff
@@ -53,7 +56,24 @@ mirai::everywhere({
   
   source(here("hardcodes.R"))
   source(here("helper_functions.R"))
+  source(here('tab_instructions.R')) # static HTML text elements
```
alex-silverman, line 59: shouldn't need this here because the mirai only deals with the 05 and 06 processing. It returns data and shouldn't need to reference (and can't update) the instruction text at all.

demo_management.R
```diff
@@ -1,13 +1,14 @@
+# This helps manage "canceling" a switch to/from demo mode
+in_demo_mode_compare <- reactiveVal(FALSE)
```
alex-silverman, line 2: why do we need this vs. just the input$in_demo_mode? Can you update the comment to explain that?

alex-silverman (changes requested):
a few small changes, and one bug found via the shinytests, but otherwise looks great! Thanks for your work on this!

global.R
mlfurman3: Agreed, will remove. I think I had added it a while back when there may have been a UI output included in one of those scripts.

PR #788, opened 2025-07-03
Dev

mlfurman3:
Contains 2 sets of updates
- metadata PR #662 
- fix for SSVF HP issue #783


alex-silverman (approved):
updated changelog date, but otherwise looks good! @mlfurman3 have we tested the metadata code, though? Just want to make sure it doesn't significantly alter the exiting metadata charts.

PR #790, opened 2025-07-09
Mbm speedups with collapse package

mlfurman3:
Convert most of the dplyr code, some data.table code in 00 through 07 scripts to collapse version. This should provide considerable speedups to the analyses based on my local testing and profiling.

Rough averages from my testing, with some instances of faster runs:
- 1 - 2 seconds for the ICF-good-small dataset
- 2 - 3 seconds for demo mode
- 10-12 seconds for main-valid
- 20-22 seconds for COHHIO

I am not sure yet how those changes will translate to the speed on the shiny server.


05_DataQuality.R
```diff
@@ -545,88 +548,116 @@ top_percents_long_stayers <- base_dq_data %>%
...
   ))) %>%
-  ungroup() %>% 
-  merge_check_info(checkIDs = 104) %>%
-  select(all_of(vars_we_want))
+  fungroup() %>%
+  fsubset(Days > quantDays) %>%
+  merge_check_info_dt(checkIDs = 104) %>%
+  fselect(vars_we_want)
 
+print(all.equal(top_percents_long_stayers2, top_percents_long_stayers, check.attributes = FALSE))
```
alex-silverman, line 592: delete

alex-silverman (changes requested):
1) Delete top_percents_long_stayers2 and the comparison

2) see if globally setting the collapse's sort=false changes overall run time in any noticeable way. You can add this in global.R in the `set_collapse` call, I think. Right now, according to the `fgroup_by` help, the default value for sort is .op[["sort"]], so it's defaulting to whatever option is set there.

global.R
```diff
@@ -28,7 +28,7 @@ library(promises)
 options(shiny.maxRequestSize = 200000000) # <- about 200MB, aka 200*1024^2
 options(shiny.fullstacktrace = TRUE)
 options(shiny.stacktraceon = TRUE)
-set_collapse(na.rm = TRUE, verbose = FALSE)
+set_collapse(na.rm = TRUE, verbose = FALSE, sort = FALSE)
```
alex-silverman, line 31: nice. yea, I guess there's no need to sort by the group, since group-based operations will still work within each group.

PR #797, opened 2025-07-22
convert date format in changelog DT table to YYYY-MM-DD

mlfurman3:
This allows the Date column to be sorted chronologically instead of alphanumerically


alex-silverman (approved):
awesome. thanks!

PR #822, opened 2025-09-15
806 new pdde and dq checks

sarasoko:
merge these new checks into monthly-inflows-outflows
I updated snapshots and csv helpers such that testthat.r runs now
- note that some snapshot changes that I approved seem to be from the sex/gender related columns now being extra columns, which was not a change made on this branch

Additionally, I left a question on the old Pull request regarding revisions to the 'Project Start Date' Issue/Guidance language. I updated the Issue & Guidance to say 'Entry Date' instead of 'Project Start Date' (leaving it alone for the data element) but did not update the 'Project End Date' replacing with 'Exit Date'.


public-resources/EvaChecks.csv
```diff
@@ -106,6 +108,7 @@ pdde,High Priority,Missing Geography Information,2.03.8 - Continuum of Care Info
 pdde,High Priority,Overlapping CE Participation Records,2.09 CE Participation Status,This project has more than one CE Participation record that covers the same time period. Please be sure you are ending any records that are no longer accurate before creating a new record.,,128
 pdde,High Priority,Overlapping HMIS Participation Records,2.08 HMIS Participation,This project has more than one HMIS Participation record that covers the same time period. Please be sure you are ending any records that are no longer accurate before creating a new record.,,131
 pdde,High Priority,Client-level data in VSP organization,2.01.3 Victim Service Provider,"Regardless of HMIS Participation Status, projects under Organizations marked as Victim Service Providers should not have client data in HMIS.",,139
+pdde,High Priority,No bed nights entered for ES - NbN project, 4.14 Bed Night,"This ES - NbN project has enrollments but no recorded bed nights. There must be a record of a bed night on the enrollment's 'Project Start Date' into a night-by-night shelter; any additional bed night dates must be after the 'Project Start Date' and before the 'Project Exit Date'.",,106
```
alex-silverman, line 111: Change this:

```This ES - NbN project has enrollments but no recorded bed nights.```

To this:

```This ES - NbN project has no enrollments with recorded bed nights.```

File: 06_PDDE_Checker.R [hunk 2]
```diff
@@ -482,6 +482,24 @@ ES_BedType_HousingType <- activeInventory %>%
...
+
+nbn_nobns <- Enrollment %>% # Get enrollments whose projects were NBN and had HMIS Participation
+  fselect(EnrollmentID, ProjectID) %>%
+  join(session$userData$Project0, on = "ProjectID", how = 'left') %>%
+  fsubset(ProjectType == es_nbn_project_type ) %>%
+  fselect(EnrollmentID, ProjectID, ProjectName, OrganizationName) %>%
+  join(HMISParticipation, on = "ProjectID", how = 'left') %>%
+  fsubset(HMISParticipationType == 1 ) %>% 
+  fselect(EnrollmentID, ProjectID, ProjectName, OrganizationName) %>%
+  unique()
```
alex-silverman, line 495: Couple things:

1. I talked with Victoria and we realized the original wording of the ticket was not precise enough. We want to flag projects where ALL its enrollments have no bed nights. This distinguishes it from the similar DQ enrollment check.
2. Use `funique()` rather than `unique()`

05_DataQuality.R
```diff
@@ -1736,7 +1791,7 @@ calculate_outstanding_referrals <- function(dq_data){
 outstanding_referrals <- calculate_outstanding_referrals(base_dq_data)
 
 # All together now --------------------------------------------------------
-dq_main <- rowbind(
+dq_main <- as.data.table(rbind(
```
alex-silverman, line 1794: You shouldn't need to convert to `data.table`, since all the underlying datasets should be `data.table`. 

But perhaps some are still `data.frame`, in which case, try `microbenchmark`ing `as.data.table()` compared to `qDT()`. The latter is a `collapse` package function and I suspect is a bit faster.

alex-silverman (changes requested):
a few small changes to make, esp. to the new PDDE check guidance text and code. (which will require updating shinytests)

06_PDDE_Checker.R
```diff
@@ -482,6 +482,24 @@ ES_BedType_HousingType <- activeInventory %>%
...
+
+nbn_nobns <- nbn_nobns %>% # anti-join gives all enrollments that have no corresponding Service records
+  join(Services, on = "EnrollmentID", how = "anti" ) %>%
+  merge_check_info(checkIDs = 106) %>% 
+  fmutate(Detail = "") %>%
+  fselect(PDDEcols) %>%
+  unique()
```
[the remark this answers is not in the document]
sarasoko: This check was made first, I believe, so the DQ dataset didn't exist yet when I wrote it. I'll update to use the DQ table

05_DataQuality.R
```diff
@@ -1474,6 +1474,54 @@ conflicting_ncbs_entry <- base_dq_data %>%
...
+    bn_eq_entry = DateProvided == EntryDate,
+    bn_eq_exit = DateProvided == ExitAdjust
+  ) %>%  
+  fgroup_by(EnrollmentID) %>% # For each enrollment,
+  fsummarise( # flag if it has any Service records where DateProvided == EntryDate/ExitAdjust
+    has_bn_eq_entry = any(bn_eq_entry, na.rm=TRUE),
+    has_bn_eq_exit = any(bn_eq_exit, na.rm=TRUE)
+  ) 
+
+missing_bn0 <- base_dq_data %>%
```
alex-silverman, line 1492: Can we rename this "nbn_w_hmis_participation" for clarity, since you use it 06_PDDE_Checker.R?

06_PDDE_Checker.R
```diff
@@ -482,6 +482,25 @@ ES_BedType_HousingType <- activeInventory %>%
...
+  ) %>% fungroup()
+
+rm(missing_bn0, services_chk)
+nbn_nobns <- nbn_nobns %>% filter(miss_all_enroll) # filter to projects with all enrollmentID missing
+
+nbn_nobns <- nbn_nobns %>% 
+  merge_check_info(checkIDs = 106) %>% 
+  fmutate(Detail = "") %>%
+  fselect(PDDEcols) %>%
+  unique()
```
alex-silverman, line 503: I think doing an `anti` join would be a lot more intuitive and concise. Like:

```
nbn_nobns <- missing_bn0 %>% # Get enrollments whose projects were NBN and had HMIS Participation
  fselect(EnrollmentID, ProjectID, ProjectName, OrganizationName) %>%
  funique() %>%
  join(Services, on = "EnrollmentID", how = "anti") %>%
  merge_check_info(checkIDs = 106) %>% 
  fmutate(Detail = "") %>%
  fselect(PDDEcols) %>%
  funique()
```

then you also don't need to persist `services_chk`.

05_DataQuality.R
```diff
@@ -1474,6 +1474,54 @@ conflicting_ncbs_entry <- base_dq_data %>%
...
+  join(services_chk, on = "EnrollmentID", how = 'anti') # EnrollmentID does NOT appear in services
+
+services_chk1 <- services_chk %>% # EnrollmentID appears in services
+  join(missing_bn0 , how="inner")  # limit to Nbn & HMISParticipationType == 1
+
+missing_bn2 <- services_chk1 %>%
+  fsubset(!has_bn_eq_entry) %>% # but it does not appear on EntryDate
+  fselect(-has_bn_eq_entry, -has_bn_eq_exit)
+
+missing_bn_entry <- missing_bn1 %>% rbind(missing_bn2) %>% as.data.table() %>%
```
alex-silverman, line 1507: I don't think we need the `as.data.table()`, since everything should be coming in as data.table now. When I tried removing those, the code still seemed to work.

06_PDDE_Checker.R
alex-silverman: You're totally right! I still feel like there's another, less roundabout way. But I can't think of one, so let's go with this. Sorry about that!

PR #825, opened 2025-09-22
New data standards

alex-silverman:


alex-silverman:
I think it's okay that the tests aren't passing. I'd rather not have to deal with updating all the libraries and stuff.

detect_invalid_values.R
```diff
@@ -0,0 +1,104 @@
+# FY26 vlaidation specs
+validation_specs_bk <- "/media/sdrive/projects/CE_Data_Toolkit/FY26 HMIS-CSV-Machine-Readable-Specifications.xlsx"
+
```
mlfurman3, line 3: I would suggest removing full filepath

PR #851, opened 2025-10-21
System overview colors and fills

mlfurman3:
Modifies color palettes and fill patterns used in Sys overview and MbM charts. Colors are listed by names in `_brand.yml`, then pulled into `eva_chart_colors.R` using the {brandr} package. Pattern fills use the {fillpattern} package.

- System Flow: Update Inflow-related categories use solid grid lines and Outflow-related ones use diagonal dashed lines
- System Status: Update to use red and blue colors for Homeless and Housed
- System Composition: Update green gradient to purple


mlfurman3:
@alex-silverman this is now ready for review

global.R
```diff
@@ -24,10 +24,14 @@ library(plotly)
 library(collapse)
 library(flextable)
 library(promises)
+library(fillpattern) # used for pattern fills in sys overview charts
```
alex-silverman: I'm torn between (a) keeping the library call here to make it clear which libraries the app uses, and (b) removing it and just calling `fillpattern::scale_fill_pattern` explicitly, to avoid having to load the whole library, since we only use it for that one function.

eva_chart_colors.R
```diff
@@ -0,0 +1,150 @@
...
+  "First-Time \nHomeless" = "grid_solid",
+  "Active at Start: Homeless" = "grid_solid",
+  "Returned from \nPermanent" = "grid_solid",
+  "Re-engaged from \nNon-Permanent" = "grid_solid",
+  "Exited, \nNon-Permanent" = "stripe120_longdash",
+  "Exited, \nPermanent" = "stripe120_longdash",
+  "Outflow" = 'stripe120_longdash',
+  "Inactive" = 'stripe120_longdash',
+  "Active at End: Housed" = 'stripe120_longdash'
+)
```
alex-silverman: similar comment to the colors re: treating this like a dictionary (and renaming to `bar_patterns` since it applies to mbm and other charts).

global.R
mlfurman3: Updated to only use :: notation, but please check that it works on your end

eva_chart_colors.R
```diff
@@ -0,0 +1,150 @@
...
+
+mbm_outflow_bar_colors <- c(
+  "Outflow" = get_brand_color('med_grey'),
+  "Active at End: Housed" = get_brand_color('dark_blue')
+)
+
+mbm_bar_colors <- c(
+  mbm_inflow_bar_colors,
+  mbm_outflow_bar_colors
+)
```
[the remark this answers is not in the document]
mlfurman3: Yes, good call these are now combined.

PR #854, opened 2025-10-27
Hotfix/additional invalid issues

alex-silverman:
### Primary Purpose
Use machine specs stored in a souped-up XLSX file on ACE, as well as some hardcoded special rules, to construct and execute basic checks on data. The specs, as used by Eva, are stored in RDS files that we must regenerate whenever the specs, EvaChecks, or the column priorities change (i.e. if we start using a column)

### Additional changes

- Added new issues and checking more datasets for these issues
- Removed some DQ checks (e.g. missing_destination_subsidy) that are now superseded by the above checks
- Added Column name detail for Impermissible Character issues in the FSA export. More precise location info is still in the Impermissible Character export.
- Changes to guidance 
- updating demo.zip to fix some of the issues flagged by the new checks
- added new test script for demo.zip in order to more quickly but fully gutcheck test new code
- Adding Exit to DQ-mirai dependencies, and Project to the PDDE ones. they are used in some of the new machine-readable specs checks, e.g. Null Unless.
- making naming conventions of Issues, Priority, etc. consistent across exports.
- Added an Empty File initially-invalid check
- Modified the columns.csv file to only be about priority. The rest is covered by the specs data.
- Updated creation of datasets to account for new, broad specs checks
- Only do trycatch on the 01-07 processing scripts when NOT in dev mode. In dev mode, we want to know if failing
- Create a `format_mirai_traceback` function for easier debugging of the `mirai`.
- Removed old EvaChecks row (id=64) for Unknown Residence prior (actual check removed here: https://github.com/abtassociates/eva/commit/d1de8d772356279d1cc8ee5018cddf5d4d50326e)

### Bug fixes

- tagging character columns as UTF-8. Strings are tagged with a certain encoding metadata that is critical for some data.table functions (e.g. `roworder`). Before, we were only handling the file encoding, not the individual strings.

### Small bug fixes

- handling nulls in Export when logging to console


mlfurman3:
@alex-silverman getting an error in `00_initially_valid_import.R` that it cannot find the `cols_and_data_types` object.

L84
```
expected_files <- unique(cols_and_data_types$CSV)
```
Warning: Error in get_expected_rclasses: object 'cols_and_data_types' not found

It happens both in demo mode and with a file upload.

alex-silverman:
> @alex-silverman getting an error in `00_initially_valid_import.R` that it cannot find the `cols_and_data_types` object.
> 
> L84
> 
> ```
> expected_files <- unique(cols_and_data_types$CSV)
> ```
> 
> Warning: Error in get_expected_rclasses: object 'cols_and_data_types' not found
> 
> It happens both in demo mode and with a file upload.

@mlfurman3 `cols_and_data_types` is now defined in `load_machine_readable_specs.R`, which is sourced by global.R. Can you confirm that you see the load script and that it's sourced in global.R? I've tried running the app from a fresh session

mlfurman3:
Yes, I see the load script in the R folder and it is called in global.R. The issue is that on L2, this if statement is TRUE but `cols_and_data_types` is only defined in the FALSE case 

`if(file.exists(specs_prepped_path)) {`

If I switch the condition to FALSE, I can run demo mode or upload a file, which is appropriately stopped by the new checks.

mlfurman3:
Last check I did here gave this error when turning on demo mode

Warning: Error in eval: object 'unique_id_lookup' not found
  97: h
  96: .handleSimpleError
  95: eval [eva/machine_readable_specs_validation.R#27]

Note for both of us: I think the same will occur with `unique_id_lookup`, `csv_join_prerequisites`, and `invalid_non_null_dynamic_lists_dt` as well.

alex-silverman:
> `cols_and_data_types`

i think this is resolved? I set `cols_and_data_types` in the TRUE case now.

alex-silverman:
> unique_id_lookup

I think this is no longer a problem:
- `unique_id_lookup` is now joined onto `reporting_info`, so we don't need it anywhere else
- the other 2 are defined in the TRUE case (where the file already existS)

mlfurman3:
Overall the changes are looking much improved and working for most scenarios.

Found one modified dataset that still broke the upload, in demo dataset mods dir:  `demo_FSACheckMods_HPFSA,zip`

vlopez0603:
In reviewing, noticed that DQ and PDDE checks for "Nulls not allowed" and "Invalid Non-Null Value" were appearing as mostly empty rows in the PDDE And DQ reports.

image

image

mlfurman3:
@alex-silverman I did some digging based on the above comment, and it seems like these two types of checks, along with others, do not have AnchorIDs in `specs_rules` that they reference to pull in the correct info. 

What is the reasoning for setting up some checks with no AnchorID?

alex-silverman:
> demo_FSACheckMods_HPFSA

Nice catch. Just pushed a fix (exiting out early from part of the code if any Key Fields are not in the dataset).

alex-silverman:
> @alex-silverman I did some digging based on the above comment, and it seems like these two types of checks, along with others, do not have AnchorIDs in `specs_rules` that they reference to pull in the correct info.
> 
> What is the reasoning for setting up some checks with no AnchorID?

@vlopez0603 - which dataset did you use and which tab were you looking at? I just tried DEMO mode and didn't see a blank org in any of the HP, Error, or Warning tabs.
@mlfurman3 - I think Victoria filled out the indicator in the specs spreadsheet for whether to include Anchor IDs or not. For some files, like Users.csv, there is no Anchor ID. But I don't remember why the other fields don't have one.

alex-silverman:
> > @alex-silverman I did some digging based on the above comment, and it seems like these two types of checks, along with others, do not have AnchorIDs in `specs_rules` that they reference to pull in the correct info.
> > What is the reasoning for setting up some checks with no AnchorID?
> 
> @vlopez0603 - which dataset did you use and which tab were you looking at? I just tried DEMO mode and didn't see a blank org in any of the HP, Error, or Warning tabs. @mlfurman3 - I think Victoria filled out the indicator in the specs spreadsheet for whether to include Anchor IDs or not. For some files, like Users.csv, there is no Anchor ID. But I don't remember why the other fields don't have one.

@vlopez0603  - I think the specs dataset hadn't been updated from my dev version to the one on eva-dev. Can you try again?

vlopez0603:
Most of the blank rows have been resolved though there's still one on the PDDE side (see screenshot below). This case is actually flagging a missing Geocode for a project.

image

I think it may be on me to update the specs sheet on ACE for at least one field. I may have missed indicating to include the AnchorID for that data element/column.

vlopez0603:
@alex-silverman / @mlfurman3 - I made minor updates to the specs spreadsheet. I noticed the Incorrect Data Type check had three columns incorrectly set as PDDE instead of FSA. The Non-Null Invalid check also had one column incorrectly set as PDDE instead of FSA. 

I'm still in the process of reviewing the other checks in the specs spreadsheet.

PR #859, opened 2025-11-03
Client count speedups

alex-silverman:


alex-silverman:
> @alex-silverman changes look ok and work as expected. I would also suggest updating the `client_level_export_details.csv` helper dataset with these changes.

fixed!

PR #893, opened 2025-11-17
handling null rows in participation datasets for overlaps in 06. This…

alex-silverman:
… should resolve some of the errors in the log, e.g.

 dq_pdde_results mirai failed with error: Error in onFulfilled(...): Error in fmutate(., PreviousCEParticipationID = lag(CEParticipationID), PreviousCEStart = lag(CEParticipationStatusStartDate), PreviousCEEnd = lag(CEParticipationStatusEndDate)): Internal error: NULL can not be subset. It is invalid for a data.table to contain a NULL column.

and

dq_pdde_results mirai failed with error: Error in onFulfilled(...): Error in gsplit(NULL, g): INTEGER() can only be applied to a 'integer', not a 'NULL'

Also adding console statements

### DIFF
--- a/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-exportTestValues.json
+++ b/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-exportTestValues.json
@@ -550,7 +550,7 @@
         [
           "Length:121        ",
           "High Priority: 5  ",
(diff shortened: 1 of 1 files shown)


mlfurman3 (approved):
Changes make sense and tests are passing. Ok to merge in!

Issue #910, opened 2025-11-25
Phase 3: Bed/Unit Inventory Counts and Utilization Calculations - Average Nightly


sarasoko:
### DESCRIPTION
These counts and calculations will underpin both a system-level and project level dashboard with a series of charts focused on average nightly bed and unit inventory counts and utilization rates for residential projects. 

Inventory and Utilization calculations are only relevant to the following project types: 0, 1, 2, 3, 8, 9, 10, 13 (subtype 2). Hardcode **project_types_w_beds** meets this criterion but needs the additional subtype restriction for project type 13.

Utilization will only be calculated for projects who have at least one period of HMIS Participation HMISParticipationType == 1) during the report.

For the system-level dashboard specifically, calculations will happen first at the project-level before numbers are aggregated up to the system level.

### DEV TASKS
In 04_initial_data_prep.R, the **EnrollmentAdjust**, **ProjectSegments**, **activeInventory**, and **HMIS_participating_projects_w_active_inv_no_overflow** data frames are useful bases for the dashboard's' calculations, though they currently lack some of the granularity needed for the filters such as dedicated bed types and inventory household types.

Additionally, the branch **utilization-ala-rminor** had old code related to project-level utilization in server.R and ui.R that may be helpful as well.

- [ ] Using **HMIS_participating_projects_w_active_inv_no_overflow** (04_initial_data_prep) as a base, get the Start and End dates for when each Project was Operating, Participating, and had Active Inventory for each of the following Participation Types:
  - HMIS Participating (HMISParticipationType == 1)
  - Not Participating (HMISParticipationType == 0)
  - Comparable Database Participating (HMISParticipationType == 2)
- [ ] In addition to the fields already included in the **HMIS_participating_projects_w_active_inv_no_overflow** data frame, ensure resulting data frame(s) in prior step have the following fields:
  - VictimServiceProvider from **project_prep** data frame
  - HousingType and TargetPopulation fields from **ProjectSegments** data frame
  - HouseholdType, ESBedType, Availability, UnitInventory, CHVetBedInventory, YouthVetBedInventory, VetBedInventory, CHYouthBedInventory,	YouthBedInventory, and CHBedInventory fields from **activeInventory** data frame
- [ ] For each relevant project, calculate the Average Nightly Beds/Units for the project. Note that these calculations will also need to be broken out by participation type, household type, etc.
  - Avg Nightly Beds = (Total Beds * Days Available During Report) / 365
[...]

PR #918, opened 2025-11-26
910 phase 3 bedunit inventory counts and utilization calculations   average nightly

sarasoko:
I made an entirely new server script here: inv_util_server that started as a copy of 'system_overview_server' in commit: ef7d04487a49c127c32a66bf07aacb6755747208

the difference viewer doesn't show up since it's new, so review the commits for how code removed from data_quality_server was adapted for tasks in #910 

this server script isn't hooked up to anything yet and still retains a lot of code from 'system_overview_server'. Since the filters will be set up similarly I tried to just replace the data manipulation reactives, leaving other pieces for future adaptation or deletion

### DIFF
--- a/08_inv_util.R
+++ b/08_inv_util.R
@@ -107,21 +107,6 @@ get_quarters <- function(){
   names(quarters) <- c("Q1", "Q2", "Q3", "Q4")
   return(quarters)
 }
-get_months <- function(){
-  lastday <- as.Date(session$userData$ReportEnd)
-  y_last <- year(lastday)
-  m_last <- month(lastday)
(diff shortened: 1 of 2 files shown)


mlfurman3 (changes requested):
Thanks, Sara. I think the parts of this that are specific to BUI look good and make sense. 

My main comment is to remove or comment out the sections that may be copied/repeated from other parts of the code, to avoid issues with duplicate IDs and logic.

- `ui.R` - the IDs for `currentProviderList` and `changelog_table` should be changed/removed
- `inv_util_server.R` - lines 1-364 and 720-end look like they are duplicated from `system_overview_server.R` and should only go in one of the two.

Issue #925, opened 2025-12-04
Client Flagged as Unknown Living Situation in EVA, Living Situation Value in CSV is Known


BrittneyBehr:
**Describe the bug**
A client was flagged in the EVA Warnings file as having an Unknown Living Situation  and Unknown Months or Times Homeless. When I review the CSV source data file, both fields have known values inputted. This should not be flagged as unknown. 

**Time of Issue**
12/4/2025 9:40am EST

**To Reproduce**
1. Run CSV
2. Upload to EVA
3. Download Full Data Quality File
4. Review CSV file with Full Data Quality file - the evaluation from EVA is incorrect according to the data uploaded.

**Expected behavior**
This client should have no warnings for Unknown Living Situation or Unknown Months or Times Homeless.

**Screenshots**

Full Data Quality File Filtered on Flagged Client
<img width="657" height="72" alt="Image" src="https://github.com/user-attachments/assets/a348cc90-1235-4c01-aa13-bc28f195c12a" />

CSV Source Data With Known Values

<img width="833" height="63" alt="Image" src="https://github.com/user-attachments/assets/0dd4d63c-c69d-4d50-bda1-4ded815c0277" />

**Desktop (please complete the following information):**
 - Operating System: Windows 10* 64-bit
 - Browser: Chrome 142 on Windows 11

desireeblakeabt:
Hi @BrittneyBehr - Thank you for your question and providing helpful screenshots. Is this flagging for one particular client or for multiple? Also, what project type is this flagging for?

BrittneyBehr:
Hello,

I do not have access to the CSV specifications due to HUD removing it from the website, but in HMIS, this person has a known value for Times Homeless:

<img width="1397" height="390" alt="Image" src="https://github.com/user-attachments/assets/be6247ab-1bd7-4f55-b3c4-7b652f9a2ccb" />

desireeblakeabt:
Yes - @BrittneyBehr. In this screenshot below for the client in project 1221, the TimesHomeless response is 9 which is reflective of 'Client prefers not to answer' which seems to differ from the screenshot recently posted from HMIS. 

<img width="833" height="63" alt="Image" src="https://github.com/user-attachments/assets/df649c41-d17f-402e-a07c-6e6b76cb4ec9" />

PR #937, opened 2025-12-18
living sitituation err should only check LivingSitutation column

sarasoko:
related to issue #925 

removes months/times homeless columns from the living situtation error (id=68). These are captured in dkr_months_times_homeless (id = 69)


mlfurman3 (approved):
Looks good, thanks!

PR #954, opened 2026-01-20
Feature/trycatch upload scripts

mlfurman3:
- Adds tryCatch blocks around the numbered upload processing scripts
- if a script encounters an error, shows a modal box with the date, time, and name of the script for easier issue creation and log ID
- currently not compatible with demo mode, since that has a `removeModal()` call in it after `process_upload`

Note: This does not address unexpected errors in the post-processing server scripts, which could be handled with the `onSessionEnded`, `onStop`, `onUnhandledError` functions.

### DIFF
--- a/upload_server.R
+++ b/upload_server.R
@@ -102,7 +102,7 @@ process_upload <- function(upload_filename, upload_filepath) {
     } else {
      
       err <- source_trycatch("07_system_overview.R")
-      if(is.null(err)) {
+      if(!is.null(err)) {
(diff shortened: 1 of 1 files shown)


alex-silverman (approved):
this is awesome and functionality testing shows it's working as intended. I created a PR (#955 ) to try and condense the code a bit and leverage the `source_trycatch` function you created. 

Other than that, just need to update shinytests, though happy to skip that since there's another branch dedicated to updating shinytests

PR #955, opened 2026-01-21
condensing trycatches

alex-silverman:
Condensing the calls to trycatch so they are handled consistently.


mlfurman3 (approved):
Nice amendment, looks like it works well with successful and unsuccessful file uploads. Approved!

PR #980, opened 2026-03-25
Updates to Links, Tab Names, Instructions

vlopez0603:
- Updated links that referenced old HMIS Specs. New link takes users to HUD Exchange page for HMIS Data Standards and HMIS CSV Specs.
- Updated tab headers on Sys and Org-level DQ pages to remove references to the number 10
- Updated DQ page instructions to align with tab name changes
- Updated "Glossary" nav option and page header to "System Performance Glossary"


demo_management.R
```diff
@@ -44,14 +44,14 @@ observeEvent(input$pageid, {
            In Demo Mode, you can view sample downloads to understand what each 
            data quality report looks like.",
            
-           "tabPDDE" = "Welcome to the Project Data page. This page
+           "tabPDDE" = "Welcome to the Project Descriptor Data Elemnts Checker page. This page
              helps users review the Project Descriptor Data Element (PDDE) data
```
mlfurman3, line 48: *Elements

vlopez0603: Great catch. Fixed!

mlfurman3:
Text changes look good, but will need to update shinytest scripts to reflect the tab name changes.

PR #991, opened 2026-05-07
Revert glossary

mlfurman3:
Small PR to go back to old version of glossary until we have system exits in place. The new version on dev has new entries that are not yet relevant.


alex-silverman (approved):
good to go once the old file is up

Issue #994, opened 2026-05-14
Incorrect Zero Utilization Warning for Programs with Multiple HMIS Participation Records


entomacy:
**Describe the bug**
PDDE DQ Report in Eva contains Zero Utilization Warnings for programs in our HMIS that have utilization during the reporting period or when a program is not participating, but participated in the past. Our team has observed that the bug occurs when a program has multiple HMIS participation status records.

**Time of Issue**
Mostly recently, Tuesday, May 5, 2026 at 9:16 AM EST, but this has been ongoing since the addition of HMIS participation fields to the Data Standards.

**To Reproduce**
Our team runs a HUD CSV Export (typically reporting period is last 90 days)
Uploads HUD CSV Export to Eva
Downloads and opens the Data Quality Project Descriptor Data Report from Eva and observe the Data tab
Some programs with Zero Utilization had clients enrolled during the reporting period

**Expected behavior**
Programs with clients enrolled during the reporting period would not have a Zero Utilization warning.

**Screenshots**

**Desktop (please complete the following information):**
 - Operating System: Version 15.6.1 (24G90)
 - Browser: Chrome Version 148.0.7778.168 (Official Build) (arm64)

<img width="987" height="129" alt="Image" src="https://github.com/user-attachments/assets/52616a76-3ef8-4df2-8ec7-ac26e48075e8" />
<img width="748" height="28" alt="Image" src="https://github.com/user-attachments/assets/9ce6f465-2c97-4500-a905-25774a7f1556" />

vlopez0603:
Thanks for submitting this issue @entomacy. We'll start investigating. In the meantime, I have a few follow up questions:

- You mentioned the typical reporting period is the last 90 days. Can you give the date range for the HMIS CSV Export that triggered this issue?
- For the Van Wert Haven of Hope program, can you confirm if there were client enrollments within its 7/1/2020 - 4/5/2023 HMIS Participating Period?

vlopez0603:
Hi @entomacy - We were able to identify the bug related to this issue and have implemented a fix in a staging site. Please let us know if you'd like to re-try uploading your file to the temporary staging site in order to see if the fix resolves your issue.

entomacy:
Hi @vlopez0603,

Apologies for the delay!

The date range for the HUD CSV Export that I used to submit this issue was 2/3/26	to 5/5/26.

I confirmed that 101 households were enrolled 7/1/2020 - 4/5/2023. (Confirmed via Clarity Program Roster report.)

[...]

PR #1011, opened 2026-06-11
Hotfix/pdde zero util

mlfurman3:
Addresses zero util bug in #994 by correctly subsetting to HMIS participating projects


vlopez0603 (approved):
Approved!

PR #1012, opened 2026-06-12
1009 update download tabs

sarasoko:
update the client level export (overview & exits) to use updated 'officeR' package function 'sheet_write_data'. 

To resolve the sheet order and file corruption issues, I added the sheets to the excel file, so I didn't need to use 'sheet_add' function. 

For the 'exits' export, I drop the 'Instructions' sheet and the 'Monthly Statuses' and 'Adjusted Enrollment' sheets instead of writing to them. 

Adding all the sheets to the xlsx template allowed me to try out different formatting to test what gets retained after writing. The column widths and conditional formatting seemed to stay, but any cell formatting (like bold and large first row) got lost. I set it to open on the 'Instructions' page


sarasoko:
closing and opening a PR into dev instead of system-exits

PR #1020, opened 2026-06-22
Timeliness CE Assessment and CE Event features

mlfurman3:
Adds Timeliness CE Assessment and CE Event records to Project Dashboard tabs and data download. Additionally, logic adjustments are made to show CE Assess, CE Event, and CLS Record counts depending on project types.


ui.R
```diff
@@ -418,7 +418,7 @@ page_navbar(
         
         nav_panel(
           title = headerTab("Timeliness"),
-          navset_card_underline(
+          navset_underline(
             id = "client_count_ti_subtabs",
```
alex-silverman: I don't love the `id` here, since Timeliness is at the same level as Client Counts but is a different tab. But nbd.

There's also only one tab. Do we expect to introduce more?

client_counts_server.R
```diff
@@ -228,27 +229,28 @@ get_clientcount_download_info <- function(orgList = unique(client_count_data_df(
...
-    select(!!keepCols, ProjectType, nlt0, n0, n1_3, n4_6, n7_10, n11p, mdn) %>%  
-    mutate(ProjectType = project_type(ProjectType)) %>% 
-    arrange(OrganizationName, ProjectName) %>% 
-    nice_names_timeliness(record_type = 'start')
+    validationStart <- tl_df_project_start() %>% 
+      fsubset(OrganizationName %in% orgList) %>% 
+      select(!!keepCols, ProjectType, nlt0, n0, n1_3, n4_6, n7_10, n11p, mdn) %>%  
+      mutate(ProjectType = project_type(ProjectType)) %>% 
+      arrange(OrganizationName, ProjectName) %>% 
+      nice_names_timeliness(record_type = 'start')
```
alex-silverman:
```suggestion
    validationStart <- tl_df_project_start() %>% 
      fsubset(OrganizationName %in% orgList) %>% 
      fselect(c(keepCols, "nlt0", "n0", "n1_3", "n4_6", "n7_10", "n11p", "mdn")) %>%  
      fmutate(ProjectType = project_type(ProjectType)) %>% 
      roworder(OrganizationName, ProjectName) %>% 
      nice_names_timeliness(record_type = 'start')
```

Also, I'd suggest putting this in a function because this same code is essentially reused below.

client_counts_server.R
alex-silverman: I think this should be a function, since it's reused a bunch below:

```
pull_time_cols <- function(cond, df) {
  if(cond)
    df %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
  else
    0
}

dat$proj_start <- pull_time_cols(
  !is.null(df) && input$currentProviderList %in% df$ProjectName,
  tl_df_project_start()
)
```

client_counts_server.R
```diff
@@ -709,16 +815,34 @@ output$timelinessTable <- renderDT({
...
-  if(cc_project_type() %in% project_types_w_cls & input$currentProviderList %in% tl_df_cls()$ProjectName){
+  if(cc_project_type() %in% c(es_nbn_project_type, out_project_type, ce_project_type)){
+    dat$cls = tl_df_cls() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
+  } else if(cc_project_type() %in% project_types_w_cls & input$currentProviderList %in% tl_df_cls()$ProjectName){
     dat$cls = tl_df_cls() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
   } else {
     dat$cls <- NULL
   }
   
+  if(cc_project_type() == ce_project_type){# & input$currentProviderList %in% tl_df_ce_assess()$ProjectName){
```
alex-silverman: just curious, why did you split up the conditions this way, as "or" separated vs. "and" like in `dat$nbn`?

client_counts_server.R
```diff
@@ -709,16 +815,34 @@ output$timelinessTable <- renderDT({
...
+    dat$ce_assess <- NULL
+  } 
+  
+  if(cc_project_type() == ce_project_type){# & input$currentProviderList %in% tl_df_ce_event()$ProjectName){
+    dat$ce_event = tl_df_ce_event() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
+  } else if(input$currentProviderList %in% tl_df_ce_event()$ProjectName){
+    dat$ce_event = tl_df_ce_event() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
+  } else {
+    dat$ce_event <- NULL
+  }
```
alex-silverman: can combine the first two conditions using `||` operator.

client_counts_server.R
alex-silverman: use "&&" for consistency

client_counts_server.R
```diff
@@ -709,16 +815,34 @@ output$timelinessTable <- renderDT({
   } else {
     dat$nbn <- NULL
   }
-  
-  if(cc_project_type() %in% project_types_w_cls & input$currentProviderList %in% tl_df_cls()$ProjectName){
+  if(cc_project_type() %in% c(es_nbn_project_type, out_project_type, ce_project_type)){
+    dat$cls = tl_df_cls() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
+  } else if(cc_project_type() %in% project_types_w_cls & input$currentProviderList %in% tl_df_cls()$ProjectName){
```
alex-silverman: can combine these conditions

alex-silverman (changes requested):
mostly formatting changes

client_counts_server.R
alex-silverman: this fails if you upload "S:\projects\CE_Data_Toolkit\Data Sets\Demo Dataset Mods\demo_ESOnly.zip" because `cc_project_type()` resolves to `NA` which isn't handled correctly.

alex-silverman (changes requested):
A few code cleanups (making things into functions), but more importantly, I'm not seeing a CLS column when I select Organization A - ES in DEMO Mode. I think it's because they have no CLS records. The ticket suggests we should still show the CLS column, but `tl_df_cls` has a join with the CurrentLivingSituation dataset, so if there are no CLS records for the selected project, the CLS column won't be shown.

client_counts_server.R
```diff
@@ -709,16 +827,34 @@ output$timelinessTable <- renderDT({
   } else {
     dat$nbn <- NULL
   }
-  
-  if(cc_project_type() %in% project_types_w_cls & input$currentProviderList %in% tl_df_cls()$ProjectName){
+  if(cc_project_type() %in% c(es_nbn_project_type, out_project_type, ce_project_type)){
+    dat$cls = tl_df_cls() %>% fsubset(ProjectName == input$currentProviderList) %>% fselect(time_cols) %>% unlist
```
alex-silverman: Ideally, I think `input$currentProviderList` would be a dropdown whose *values* are Project IDs and whose display names are Project Name, and then use Project ID for subsetting, just in case there are duplicate project names. But def nbd if it's hard to do.

Also, the ticket suggests we should show the CLS column for NbN projects, no matter what, but `tl_df_cls()` has a join with the CurrentLivingSituation dataset, so if there are no CLS records for the selected project, the CLS column won't be shown. Maybe that's desirable, but just bringing it up since it's discrepant with the ticket.

client_counts_server.R
alex-silverman:
```suggestion
    how = "inner"
```

I think we want `inner` because on the next line, we subset to non-null `CurrentLivingSituation.DateCreated`, which is only possible if there's a matching CLS record.

alex-silverman:
Also, when I run main-valid in this branch, the following files have changed:
- ~/eva/tests/helper_data/main-valid/dq_main.csv
- ~/eva/tests/helper_data/main-valid/pdde_main.csv
- ~/eva/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-exportTestValues.json
- ~/eva/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-just-uploaded.json

But I don't see these in the PR. `dq_main.csv` lost 8 records, 4 each of: Missing Current Living Situation Subsidy Type and Missing Prior Living Situation Subsidy Type. `pdde_main.csv` lost 2 records (issues Inventory Start Precedes Operating Start and Incorrect CoC in bed inventory) and gained 1 (issue No Inventory Records).

www/custom.css
```diff
@@ -271,6 +271,10 @@ body.idle-timeout-occurred #ss-connect-dialog {
   border: 3px solid #FCB248;
 }
 
+#client_count_ti_subtabs {
```
alex-silverman: i think this should be `project_dashboard_tl_subtabs`. That's the id in `ui.R`.

alex-silverman (changes requested):
one super quick change in custom.css. Otherwise looks good for merge!

project_dashboard_server.R
alex-silverman, line 1: - I would rename this file to `project_dashboard_server.R`
- I'd found a few other name changes (e.g. `downloadClientCountReport` and related elements)

project_dashboard_server.R
```diff
@@ -573,12 +594,10 @@ tl_df_cls <- reactive({
   ## Time to Entry - CLS
   cls_df <- join(
     client_count_data_df(), 
-      #fsubset(ProjectName == input$currentProviderList) %>% 
     session$userData$CurrentLivingSituation %>% 
       fselect(PersonalID, EnrollmentID, CurrentLivingSituation.DateCreated = DateCreated, CurrentLivingSituation.InformationDate = InformationDate),
```
alex-silverman: @mlfurman3 - why do we prefix `DateCreated` and `InformationDate` with "CurrentLivingSituation"?

client_counts_server.R
alex-silverman: @mlfurman3 bringing this back up, jic it's unintentional.

project_dashboard_server.R
```diff
@@ -682,43 +772,80 @@ output$timeliness_vb3 <- renderUI({
...
+      dat$cls = NULL
+    }
   } else {
-    dat$nbn <- NULL
+    dat$cls <- pull_time_cols(
+      cc_project_type() %in% project_types_w_cls && input$currentProviderList %in% tl_df_cls()$ProjectID,
+      tl_df_cls(),
+      set_zero = FALSE
+    )  
   }
```
alex-silverman, line 822: I think I'm missing something, but why not combine these?

```Suggestion
dat$cls <- pull_time_cols(
    cc_project_type() %in% c(es_nbn_project_type, out_project_type, ce_project_type) || (cc_project_type() %in% project_types_w_cls && input$currentProviderList %in% tl_df_cls()$ProjectID),
    tl_df_cls(),
    set_zero = FALSE
  )  
```

alex-silverman (commented):
a few clarifying questions

client_counts_server.R
mlfurman3: Corrected this, thanks

client_counts_server.R
mlfurman3: Combined, thanks

client_counts_server.R
mlfurman3: Added this function!

project_dashboard_server.R
mlfurman3: I had added this for all the timeliness dfs at first to keep track of which data frame the date fields came from. But not really needed anymore, so fixed it across the board.

project_dashboard_server.R
mlfurman3: I think I kept them separate because of this ask - wanted to preserve the 0s for ES, SO, and CE projects but not others. So the two conditions aren't really the same. 

image

But I could be mistaken and there is an easy way to combine them.

alex-silverman: I just don't see the difference. In both cases:
- if the project type is one of the ones in the list 
- the selected project ID is in the dataset
- then we select `time_cols`
- and unlist
- otherwise we return NULL

I guess I don't understand:
1. Where does the CurrentLivingSituation column get added for either case? `time_cols` doesn't contain that column name, so in either case we're selecting `time_cols` and not including that name. It's not even in `tl_df_cls`.
2. what is the point of the `input$currentProviderList %in% tl_df_cls()$ProjectID` types of statements? Isn't that what the `fsubset` is for?

alex-silverman: okay, I think I understand the code a bit better now, but I still think there are some scenarios that won't work right:

- CLS column is hidden when ES-NbN, SO, or CE projects have zero records
- CE Assessment and CE Event columns are hidden for CE projects with zero records
- cond is TRUE, but the project has zero actual records of that type. This will result in 0-length vector, which might crash the app when trying to be rendered

project_dashboard_server.R
```diff
@@ -521,14 +545,13 @@ tl_df_project_exit <- reactive({
   ## Time to Entry - Project Exit
   df_exit <- join(
     client_count_data_df() %>% 
-      #filter(ProjectName == input$currentProviderList) %>% 
       rename(ProjectExitDate = ExitDate),
```
alex-silverman: Why do we rename to ProjectExitDate instead of leaving as ExitDate?

project_dashboard_server.R
```diff
@@ -521,14 +545,13 @@ tl_df_project_exit <- reactive({
...
-    session$userData$Exit %>% fselect(PersonalID, EnrollmentID, Exit.DateCreated = DateCreated),
+    session$userData$Exit %>% fselect(PersonalID, EnrollmentID, DateCreated),
     how = "left"
   ) %>%  
     fsubset(between(ProjectExitDate, input$dateRangeCount[1], input$dateRangeCount[2])) %>% 
-    fmutate(DaysToEntry = as.numeric(ProjectExitDate - as.Date(Exit.DateCreated) ),
-            HoursToEntry = as.numeric(difftime(ProjectExitDate, Exit.DateCreated, units="hours"))) %>% 
+    fmutate(DaysToEntry = as.numeric(ProjectExitDate - as.Date(DateCreated) ),
```
alex-silverman: - Are these inverted? Should this be DateCreated - ProjectExitDate?

project_dashboard_server.R
```diff
@@ -521,14 +545,13 @@ tl_df_project_exit <- reactive({
...
+            HoursToEntry = as.numeric(difftime(ProjectExitDate, DateCreated, units="hours"))) %>%
```
alex-silverman: seems like we don't use HoursToEntry anymore.

tab_instructions.R
alex-silverman: "comapring" is misspelled

alex-silverman (changes requested):
A couple potential issues:
- I think there are some edge scenarios that will cause bugs in the `output$timelinessTable` renderDT, in the creation of `dat$cls` and `dat$ce`. Left a comment in the place I mean
- I think the DaysToEntry is inverted for the `tl_df_project_exit`

In general, I think there is a lot of redundant code that could be condensed/turned into functions. Not something we need to fix now, but going forward, if you find yourself repeating code, make a function instead. It makes the code more readable and updateable.

project_dashboard_server.R
mlfurman3: No longer needed, but this was to help us differentiate between Exit$ExitDate and Exit$DateCreated. Will remove

tab_instructions.R
mlfurman3: Fixed, thanks!

project_dashboard_server.R
mlfurman3: You are correct, I have removed it

project_dashboard_server.R
mlfurman3: yes, this was backwards - thank you for catching!

alex-silverman (approved):
pushed some changes to timeliness_cleanup based on AI review pointing out potential problems with null handling

PR #1022, opened 2026-07-01
Hotfix/fix missing dest updates

alex-silverman:
fix bug flagged by shinytest where ReferenceNo is not the same type as the hardcoded fixes, causing the `fcase` in `fix_missing_destination` to crash the app.


mlfurman3 (approved):
Works well, thanks

Issue #1032, opened 2026-07-14
Bug: System Inflow/Outflow Data Download


sarasoko:
I noticed this in the Eva Prod shiny log:

```
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.264988  Inflow/Outflow data download
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.266288  In sys_export_monthly_info
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.266971  In sys_monthly_chart_data_wide
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.267578  In sys_inflow_outflow_monthly_chart_data
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.268091  In get_inflow_outflow_monthly
27929eb091243be167c60d8272adcd6e  2026-07-07 11:08:19.2684  In get_inflow_outflow_monthly, num months_data records: 3807
Warning: Error in : Error in libxlsxwriter: 'String exceeds Excel's limit of 32,767 characters.'
```

Not sure how to replicate, but there's presumably some cell that contains more than 32,767 characters. We would need to truncate. Not sure which column is most likely.

Issue #1030, opened 2026-07-14
Bug: DQ Download


sarasoko:
I noticed this error in the latest 6/15/2026 shiny log on Eva Prod.

```
77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.052324  in dqDownloadInfo
77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.053189  in dq_full
77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.056606  in getDQReportDataList
77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.164987  in getDQReportDataList
77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.502859  in get_Dq_plot_data, level = org, issueType = High Priority, groupVars = OrganizationName77991ceb5c5d7b5119c24dba0c4ae5b7  2026-07-03 09:06:08.502859  in get_Dq_plot_data, level = org, issueType = High Priority, groupVars = Issue
Warning in file(con, "rb") :
  'raw = FALSE' but '/srv/shiny-server/eva/www/js' is not a regular file
Warning in file(con, "rb") :
  cannot open file '/srv/shiny-server/eva/www/js': it is a directory
Warning: Error in file: cannot open the connection
...
Warning: Error in which: argument to 'which' is not logical
```

Given how many instances of this same error I see in a row, my suspicion is that it's happening somewhere in `output$dq_export_download_btn`, specifically, this line:

```
for(i in orgs_to_save){
          dq_export_list <- get_dqDownloadInfo_export(i, value = "org")
          progress$inc(amount = 1 / length(orgs_to_save),
                       detail = paste0('Org ', which(orgs_to_save == i), ' of ', length(orgs_to_save)))          # <-- HERE
         ...
}
```

But I couldn't replicate this.

sarasoko:
@alex-silverman 

I also can't replicate this error but I made the following changes to data_quality_server.R on mt linked branch '1030-bug-dq-download':

1 - fixed the console Log for "in get_Dq_plot_data" to use {paste(groupVars, collapse = ', ')} instead of {groupVars} so that the message isn't duplicated

2 - eliminated all uses of which() by adjusting the loop to be for i in 1:length(orgs_to_save) instead of for i in orgs_to_save

I'm not entirely sure why the following warnings are occurring and could use some help.
[...]

PR #1034, opened 2026-07-15
1030 bug dq download

sarasoko:
Implementing fixes for two issues #1030 & #1032 on the same branch.

I originally branched of 'main' but merged in dev since it was missing edits to how the client level export works.

I've addressed the 'which' error and the value exceeding max character length in xlsxwriter. 

I'm still not entirely sure why the following warnings are occurring and could use some help.
Could these errors have just been caused by the other error with 'which'?

```
Warning in file(con, "rb") :
  'raw = FALSE' but '/srv/shiny-server/eva/www/js' is not a regular file
Warning in file(con, "rb") :
  cannot open file '/srv/shiny-server/eva/www/js': it is a directory
Warning: Error in file: cannot open the connection
```
Perhaps they are caused by the button toggle for valid_file()?

```
observe({
  
  if('Organization-level (multi-select)' %in% input$dq_export_export_types){
    shinyjs::show(id = 'dq_export_orgList')
  } else {
    shinyjs::hide(id = 'dq_export_orgList')
  }
  
  shinyjs::toggle("dq_export_download_btn", 
                  condition = (session$userData$valid_file() == 1))
                  #condition = (session$userData$dq_pdde_mirai_complete() == 1))
})

```
or the enable/disable?

```
observe({
  req(session$userData$valid_file() == 1)
  #req(session$userData$dq_pdde_mirai_complete() == 1)
  ## disable DQ Export button if any of these cases are true
  if(
    ## (1) org-level and system-level are both unchecked
    length(input$dq_export_export_types) == 0 | 
    ## (2) org-level is selected but all orgs are de-selected
     ("Organization-level (multi-select)" %in% input$dq_export_export_types & length(input$dq_export_orgList) == 0) | 
    ## (3) one or both checkboxes are checked, but all report types are unchecked
    (length(input$dq_export_files) == 0 )
  ){
    shinyjs::disable('dq_export_download_btn')
  } else {
    shinyjs::enable('dq_export_download_btn')
  }
})
```
I think both would be re-run every time the page re-freshes (to update increments) and that's why there are so many?


alex-silverman:
> I originally branched of 'main' but merged in dev since it was missing edits to how the client level export works.

i would always branch off `dev`. `dev` is either equivalent or more up-to-date than `main`.

helper_functions.R
```diff
@@ -765,6 +765,18 @@ calc_pct_change <- function(count_prev, count_current, accuracy = 1, format='cha
...
+xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = ""){
+  # checking if any columns need to be truncated
+  trunc_cols <- map(df, ~ifelse(is.character(.x), max(nchar(.x), na.rm=TRUE)>max_nchar, FALSE) ) %>% unlist()
+  
+  if(any(trunc_cols)){
+    df <- map_at(df, which(trunc_cols), ~str_sub(.x, 1, max_nchar)) %>% as.data.frame()
+    logToConsole(session, paste("truncating", log_loc, "column(s) :", paste(names(which(trunc_cols)), sep = ", "), 
+                                "that exceed the maximum number of characters", max_nchar))
+  }
+  return(df)
```
alex-silverman:
```suggestion
  df_chars <- char_vars(df)
  if(length(df_chars) == 0) return(df)
  
  trunc_flags <- df_chars |> map_lgl(~any(stringi::stri_length(.x) > max_nchar, na.rm = TRUE))
  if (!any(trunc_flags)) return(df)
  
  truncated_list <- modify_at(df_chars, which(trunc_flags), ~stringi::stri_sub(.x, 1, max_nchar))
  df_final <- ftransform(df, truncated_list)
  
  logToConsole(session, paste("truncating", log_loc, "column(s) :", paste(names(which(trunc_flags)), collapse = ", "),
                                "that exceed the maximum number of characters", max_nchar))
  return(df_final)
```

This has a few improvements: 
- avoid warnings: If a column is entirely NA, `max(..., na.rm=TRUE)` returns -Inf and prints a warning: "no non-missing arguments to max; returning -Inf". 
- speed: focuses only on char columns, and then those that need to be truncated, and avoids conversions to lists and dataframes
- logging: `collapse` instead of `sep` in the paste so that the columns come out on one line.

In general, `stringi` is faster than `stringr`, and `collapse` is often faster than base R. e.g. instead of `as.data.frame()`, doing `qDF()`

helper_functions.R
```diff
@@ -765,6 +765,18 @@ calc_pct_change <- function(count_prev, count_current, accuracy = 1, format='cha
   }
 }
 
+# truncate columns of dataframes being exported to xlsx
+xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = ""){
```
alex-silverman: need to pass in the `session` var so that `logToConsole` works

alex-silverman (changes requested):
we should submit an issue to the `officer` and `writexl` github repos for this!

I actually think it's worth rethinking where we implement this truncation function. Instead of for each dataset that we export, let's just truncate every csv when we import. In the FY26 specs, we're going to have a check on string-length for every string variable, and the max length is 500 characters, so if it's more than that, it will get flagged anyway.

helper_functions.R
sarasoko: updated in latest commits

helper_functions.R
sarasoko: updated in latest commits

helper_functions.R
```diff
@@ -766,15 +766,19 @@ calc_pct_change <- function(count_prev, count_current, accuracy = 1, format='cha
 }
 
 # truncate columns of dataframes being exported to xlsx
-xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = ""){
+xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = "", sesh = session){
```
alex-silverman: It's better practice and more consistent with the rest of our code to name the parameter `session`. Also, defaulting to the value `session` won't work in this context because `helper_functions.R` is global (called from `global.R`, but `session` is a server-level variable).

```suggestion
xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = "", session){
```

alex-silverman (changes requested):
I still think it would be better to just call `xlsx_char_trunc` from `importFile` on every csv. That way, there's only one place in the code that we call it and we ensure we catch everything. 

Is there a reason you think it's better to call it in these different places?

sarasoko:
> I still think it would be better to just call `xlsx_char_trunc` from `importFile` on every csv. That way, there's only one place in the code that we call it and we ensure we catch everything.
> 
> Is there a reason you think it's better to call it in these different places?

@alex-silverman My reasoning was that the original imports should be kept raw so that we aren't truncating before performing an evacheck on the string length. However, you said the check would be 500 characters which is much smaller than the excel limit. So, I can make this change.

helper_functions.R
```diff
@@ -217,7 +217,7 @@ importFile <- function(upload_filepath = NULL, csvFile, guess_max = 1000, sessio
   }
 
   # truncate strings exceeding the maximum excel limit
-  data <- xlsx_char_trunc(data, log_loc = paste(filename, "(importFile)"), session = session)
+  data <- xlsx_char_trunc(data, log_loc = paste(csvFile, "(importFile)"), session = session)
```
alex-silverman: move this to after the `convert_data_to_utf8` call.

helper_functions.R
```diff
@@ -774,13 +774,13 @@ xlsx_char_trunc <- function(df, max_nchar = 32767, log_loc = "", session){
   df_chars <- char_vars(df)
   if(length(df_chars) == 0) return(df)
   
-  trunc_flags <- df_chars |> map_lgl(~any(stringi::stri_length(.x) > max_nchar, na.rm = TRUE))
+  trunc_flags <- df_chars |> map_lgl(~any(stringi::stri_length(str_conv( .x , encoding = "UTF8")) > max_nchar, na.rm = TRUE))
```
alex-silverman: this change shouldn't be necessary if you move the call to `xlsx_char_trunc` to after the convert_to_utf8 code (see above)

helper_functions.R
```diff
@@ -164,7 +164,7 @@ parseDate <- function(datevar) {
   return(newDatevar)
 }
 
-importFile <- function(upload_filepath = NULL, csvFile, guess_max = 1000) {
+importFile <- function(upload_filepath = NULL, csvFile, guess_max = 1000, session) {
```
alex-silverman, line 167: I see why you did this. The problem is that we call `importFile` in a number of places, so either you'll have to pass `session` in all those places or you can stop passing session entirely and use `capture.output` to print the message, though I'm not 100% confident that shows up in the logs. I'd lean toward passing session everywhere we call `importFile`

client_level_export_server.R
alex-silverman, line 1: I think the plan was to add a call to `xlsx_char_trunc` or something else that checks whether something is getting truncated before export.

data_quality_server.R
alex-silverman, line 1: before making the changes to the way we loop, can you ask the user who reported the DQ Download issue to test the new code in a staging app to see if that fixes the problem? I'd prefer to know it fixes the problem before making the change.

alex-silverman (changes requested):
need to pass `session` to all calls to `importFile` and can we try to confirm the DQ Download fix actually fixes the problem?

data_quality_server.R
sarasoko: I added running xlsx_char_trunc to the export for the client level download (System Performance)

- I'm not sure how to tell who the original user is. I believe you just found the errors in the log and made an issue, instead of it being reported by a user
- I'm not set up to deploy to staging apps

sarasoko: https://github.com/abtassociates/eva/issues/1032

alex-silverman: @sarasoko You're right. My bad. I forgot. Can we move the changes related to the DQ Download issue to a new branch and hold off on merging for now until we get a better handle on the issue and/or can replicate it?

Once you remove those changes from this PR, let's try and get you set up to deploy to staging and deploy this change to staging app and ask the user who had reported the Client Level Download issue to test there

alex-silverman (changes requested):
As discussed, let's pull out the truncation code into a separate branch/PR, since we haven't been able to replicate nor has any user reported the DQ Download issue.

Code comments and outline-syntax is great. 

One more question: While I don't think it's a problem, can you briefly explain why you moved around the `session$userData$Services`, `session$userData$Enrollment` and `session$userData$Project0` code in the 04_initial_data_prep.R script?

sarasoko:
> As discussed, let's pull out the truncation code into a separate branch/PR, since we haven't been able to replicate nor has any user reported the DQ Download issue.
> 
> Code comments and outline-syntax is great.
> 
> One more question: While I don't think it's a problem, can you briefly explain why you moved around the `session$userData$Services`, `session$userData$Enrollment` and `session$userData$Project0` code in the 04_initial_data_prep.R script?

I did that when reorganizing so that each object is defined/set in it's own heading level (#Project0, #Enrollment, #Services)

sarasoko:
Closed and opened pull request https://github.com/abtassociates/eva/pull/1052

Issue #1031, opened 2026-07-14
"Bed Night Entered on Exit" error appears for enrollments with no bed nights


derekferreeica-collab:
**Description**

We've noticed an issue with the "Bed Night Entered on Exit" error in the EVA HUD Tool.

The error is being generated for client enrollments that have no bed nights recorded at all. This is confusing for users because they are instructed to look for a bed night on the exit date to resolve the error, but no bed nights exist anywhere in the enrollment.

In these cases, the enrollment is also frequently flagged with the "No Bed Night on Entry Date" error, which is expected since no bed nights have been recorded.

While correcting the missing entry bed night will typically resolve both errors, the "Bed Night Entered on Exit" error is still misleading in these scenarios because its wording and guidance imply that a bed night exists on the exit date when, in reality, none exist.

**Expected behavior**:

The "Bed Night Entered on Exit" error should only appear when there is actually a bed night recorded on the exit date that violates the validation logic.
If the enrollment has no bed nights at all, the "No Bed Night on Entry Date" error appears to be the appropriate validation, and the exit-related error should not be displayed.

Could you take a look into this behavior?

alex-silverman:
Hi @derekferreeica-collab - this would definitely be surprising if there are indeed no bed nights for these particular enrollments. Would you be able to confirm that **in the Services.csv file of your HMIS upload there truly are no records (with non-null DateProvided) for these enrollment IDs**? I'm struggling to think of a way that Eva would flag this "Bed night entered on Project Exit Date" error if there were no such records

derekferreeica-collab:
Hello @alex-silverman , 

Apologies for the late reply. I signed up for Github using my Gmail, which isn’t actually my primary work email. 

I just did a spot check of about 10 different records with that error cross referencing against the services file and can confirm that the clients in question are not present at all. There are 29000 service records in the file, just not for the clients with _Bed night entered on Project Exit Date_ errors. 

Is there anything else you’d like me to check out? I’d also be happy to send some files or do a quick call if that’s helpful. Our system uses Clarity by the way. 

Thanks, 
Derek

alex-silverman:
Thanks, @derekferreeica-collab . I have found the issue and am working on a fix!

vlopez0603:
Hi @derekferreeica-collab - A fix for this issue has been implemented in Eva. Whenever you have a chance, please let us know if you continue to see ES - NbN enrollments with no bed nights erroneously flagged with the "Bed Night Entered on Exit" check.

[...]

PR #1036, opened 2026-07-23
Hotfix/services checks

alex-silverman:
Fixes https://github.com/abtassociates/eva/issues/1031 by removing enrollments with no Bed Nights/Service records


vlopez0603 (approved):
Change to check looks good. A spot check of dq_main.csv seems to show that ES - NbN clients with no bed nights are dropped from this check--which is what we want.

PR #1037, opened 2026-07-27
Household Type as selector, not Filter in Exits by Subpopulation

alex-silverman:
, visually group Household Type with the other chart selectors (since it already behaves as one, rather than a filter)


mlfurman3 (changes requested):
Need to hook up new `syse_subpop_hh_type` to server-side logic in `system_exits_server.R` and `syse_subpop_server.R`. Atm the app only uses the old one and does not respond to changes in the new one.

alex-silverman:
> Need to hook up new `syse_subpop_hh_type` to server-side logic in `system_exits_server.R` and `syse_subpop_server.R`. Atm the app only uses the old one and does not respond to changes in the new one.

Duh. Oops. Just pushed a change.

mlfurman3:
I think there are still some inputID references that are off/mislabeled. In DEMO mode on main/the live site, if you change the HH type filter to "Adult Only", you see this split 

image

But for this PR branch, it shows the "No data to show" message.

system_exits_server.R
```diff
@@ -185,34 +185,42 @@ all_filtered_syse_demog <- reactive({
...
     get_system_exits(
-      enrl_data, # Pass cached reactive data frame directly
+      enrl_filtered,
       session$userData$client_categories, 
       ctype = 'exits_types',
       reportStart = session$userData$ReportStart,
       reportEnd = session$userData$ReportEnd,
       append_client_info = TRUE,
       client_categories = session$userData$client_categories
-    ) %>%
```
mlfurman3, line 198: Did we lose this join during the functionalizing? I am getting an error that HouseholdType column is not available

Image

alex-silverman: I've pushed a fix. I think this is a result of the HouseholdType not being "a filter" and therefore, yes, not being applied earlier.

mlfurman3 (approved):
Looks good to go now, thanks!

PR #1038, opened 2026-07-28
Reducing repetitive code by making more functions.

alex-silverman:


mlfurman3 (commented):
These look good and provide a much needed conciseness to the repeated steps. 

Only comment is that bc you were right that Project Exit dates were backwards, we don't need to include the step multiplying that by -1

mlfurman3:
Victoria also pointed out a bug when changing the Project Dashboard dateRangeInput that causes an error.

image

Let's add a validation check for the date range

image

alex-silverman:
> Victoria also pointed out a bug when changing the Project Dashboard dateRangeInput that causes an error.
> 
> image
> Let's add a validation check for the date range
> 
> image

I pushed fixes for these

mlfurman3 (approved):
Both fixes working as expected, thanks! Good to merge in.

PR #1049, opened 2026-08-05
fix formatting of old changelog entriesm, linking to resolved issues

alex-silverman:
also add new changelog entries for upcoming push


vlopez0603 (approved):
looks good

PR #1052, opened 2026-08-12
1032 bug system inflowoutflow data download

sarasoko:
New pull request replacing https://github.com/abtassociates/eva/pull/1034 

Isolate changes for system inflowoutflow data download bug


alex-silverman (commented):
@sarasoko In general, this works. Next step is to put this branch on a staging app and ask the user from https://github.com/abtassociates/eva/issues/1032 to try and upload there and see if we can tell what's going on. The expectation is that our code isn't handling their data correctly. But let's try to figure out where. This will be a good opportunity to get you more familiar with the server and deploying staging apps.

A couple other comments: 
- In general, we should try to reduce the number of differences viz. the PR unless there's a good reason for them. It makes reviews a little harder because it can be unclear what has changed and why. For example, the code initializing `session$userData$Project0` has been rearranged, and it's not immediately clear what has changed and why.
- We should try to keep a consistent style in terms of where we add line breaks. Of course it's not realistic to be perfectly consistent, but it's a good goal to have.

PR #1055, opened 2026-08-17
Revert "Revert "Feat/project performance metrics""

alex-silverman:
Reverts abtassociates/eva#1054


mlfurman3 (approved):
These look great and work well, I will do some UI testing for efficiency vs a conditionalPanel approach but don't want that to hold up merging in since we have so many PRs piling up.

PR #1057, opened 2026-08-19
Feat/sys perf export interface

alex-silverman:
Creating a more unified export interface for System Overview and a similar one for System Exits. There is now a System Performance Exports script that references `hardcodes.R` config lists.

Part of this required merging much of the similar code between them. For example, the Demographics stuff is quite similar between the two, so much of that was moved into a separate System Performance Demographics script. 

This also led to some additional steps to consolidate some of the code within the System Exits. 

The other thing I did was, for the shinytests, added some code to try and release some of the used memory, rather than letting the tests compound the memory. This should help us avoid dev server crashes


mlfurman3:
One issue I was seeing the other day was the download failing when checking the all boxes before a subpop was defined.

Steps to reproduce:
- turn on demo mode
- navigate to system exits tab, system exits by type subtab
- open SP export dropdown
- click the all checkboxes for PPT and data downloads (this auto-selects the subpop ones below, even though they are disabled)
- click to download

alex-silverman:
> One issue I was seeing the other day was the download failing when checking the all boxes before a subpop was defined.
> 
> Steps to reproduce:
> 
> * turn on demo mode
> * navigate to system exits tab, system exits by type subtab
> * open SP export dropdown
> * click the all checkboxes for PPT and data downloads (this auto-selects the subpop ones below, even though they are disabled)
> * click to download

good catch. Pushed a fix.

mlfurman3:
This now works in the default case, but not another.

Steps to reproduce:
- Uncheck one of the select all checkboxes
- Recheck, the box which now auto=selects the disabled subpop chart
- Attempt download

alex-silverman:
> This now works in the default case, but not another.
> 
> Steps to reproduce:
> 
> * Uncheck one of the select all checkboxes
> * Recheck, the box which now auto=selects the disabled subpop chart
> * Attempt download

Good catch. just pushed a fix.

mlfurman3:
As we discussed on tech team call, subpop box should be automatically checked once a subpop is chosen (i.e. selecting Adult Only in the HH type dropdown)

image

PR #1062, opened 2026-08-26
Enrollment Crossing OperatingEndDate

sarasoko:
# 04_initial_data_prep.R
- include OperatingEndDate in 'enrollment_positions' (of 05) by selecting it in EnrollmentOutside2

# 05_data_quality
## enrollment_x_operating_end (evacheck 119)
- update to additionally check that OperatingEndDate is not NA and that it is <= meta_HUDCSV_Export_End

# server_00_changelog
- update change log


alex-silverman (changes requested):
1. Let's fix the changelog to just show this new change for 9/17, using the new chnagelog md approach,
2. Instead of adding this extra variable and extra subsetting conditions to determine enrollment_x_operating_end, let's modify more upstream, i.e. in `Enrollmentvs`, we can add these 2 conditions when calculating "Enrollment Crosses ... End". (We could also do this for "Enrollment Crosses ... Period", but this check is already duplicative with the other two "Enrollment Crosses..." checks, so we probably should get rid of it anyway. I think I might already do that in another branch, though, so no need to do it here).

sarasoko:
> 1. Let's fix the changelog to just show this new change for 9/17, using the new chnagelog md approach,
> 2. Instead of adding this extra variable and extra subsetting conditions to determine enrollment_x_operating_end, let's modify more upstream, i.e. in `Enrollmentvs`, we can add these 2 conditions when calculating "Enrollment Crosses ... End". (We could also do this for "Enrollment Crosses ... Period", but this check is already duplicative with the other two "Enrollment Crosses..." checks, so we probably should get rid of it anyway. I think I might already do that in another branch, though, so no need to do it here).

@alex-silverman 
1. I merged change log method from dev and readded the new change log message. 
2. I reversed the changes I had made to 04/05 scripts and implemented the new logic in 'Enrollmentvs' function (04). Please note that the change will now impact 'EnrollmentvOperating' in addition to 'EnrollmentvParticipating'. 

I didn't implement this logic for any other conditions except for 'Crosses End'. However, ComparisonEnd is used for :
the 'Inside' condition (ComparisonEnd > Sys.Date or ExitAdjust <= ComparisonEnd) and (EntryDate >= ComparisonStart), 
the 'After' condition (EntryDate > ComparisonEnd),
the 'Crossed Period' condition (ExitAdjust > ComparisonEnd and EntryDate < ComparisonStart),

Should I add the condition : !is.na(ComparisonEnd) & ComparisonEnd >= Export Start to each? I think it makes sense for 'After' and 'Crosses Period' but I am unsure about 'Inside'.

Which are removed in another branch? Are the conditions removed from Enrollmentvs or just the eva checks on them?

sarasoko:
@alex-silverman Note that the shinytest changes were merged from 1061 into dev. See this commit:
dcb484ae4e23e793910117ffc49319e3b2468fda

I think when you use the 'resolve conflicts' functionality for the shiny tests, it ignores those differences with 'dev' when it runs the PR shiny tests (allowing the tests to pass on #1064). Then, when it was pulled into this branch, the shiny tests fail when it encounters the difference and does not fall back on the conflicts being "resolved" 

I would have assumed the shiny tests for 1064 would have failed, but they did not.

rcode/05_data_quality.R
```diff
@@ -1067,7 +1067,9 @@ enrollment_before_operating_period <- enrollment_positions %>%
   fselect(vars_we_want)
 
 enrollment_x_operating_end <- enrollment_positions %>%
-  fsubset(EnrollmentvOperating == "Enrollment Crosses Operating End") %>%
+  fsubset(EnrollmentvOperating == "Enrollment Crosses Operating End" & 
+            !is.na(OperatingEndDate) &
```
alex-silverman: i don't think we need this anymore, since it's captured by "Enrollment Crosses Operating End"

rcode/05_data_quality.R
```diff
@@ -1067,7 +1067,9 @@ enrollment_before_operating_period <- enrollment_positions %>%
   fselect(vars_we_want)
 
 enrollment_x_operating_end <- enrollment_positions %>%
-  fsubset(EnrollmentvOperating == "Enrollment Crosses Operating End") %>%
+  fsubset(EnrollmentvOperating == "Enrollment Crosses Operating End" & 
+            !is.na(OperatingEndDate) & 
+            OperatingEndDate <= session$userData$meta_HUDCSV_Export_End) %>%
```
alex-silverman: just confirming we want this line and we do NOT want a similar condition for "Enrollment Crosses Participating End"?

rcode/04_initial_data_prep.R
```diff
@@ -168,7 +168,7 @@ Enrollmentvs <- function(EntryDate, ExitAdjust, ComparisonStart, ComparisonEnd,
       paste0("Enrollment After ", comparisonWord," Period"),
     ExitAdjust < ComparisonStart,
       paste0("Enrollment Before ", comparisonWord, " Period"),
-    EntryDate > ComparisonStart & ExitAdjust > ComparisonEnd,
+    EntryDate > ComparisonStart & ExitAdjust > ComparisonEnd & !is.na(ComparisonEnd) & ComparisonEnd >= session$userData$meta_HUDCSV_Export_Start,
```
alex-silverman: I'm just realizing this still might not be right if both EntryDate AND ExitAdjust are after ComparisonEnd. I wonder if it would be more clear to write:

```suggestion:
data.table::between(ComparisonEnd, EntyrDate, ExitAdjust, incbounds = FALSE) & !is.na(ComparisonEnd) & ComparisonEnd >= session$userData$meta_HUDCSV_Export_Start
```

Whether we use `between` or not, when we say Crosses Start, we look at whether ComparisonStart falls between EntryDate and ExitAdjust, so I'd think we'd want to do the same for Crosses End

alex-silverman:
> 1. I merged change log method from dev and readded the new change log message.
> 2. I reversed the changes I had made to 04/05 scripts and implemented the new logic in 'Enrollmentvs' function (04). Please note that the change will now impact 'EnrollmentvOperating' in addition to 'EnrollmentvParticipating'.

I think this is a good change, but let's discuss in next dev meeting. I believe that, without this change, if the comparison end date is NA, then the whole thing resolves to NA or maybe errors out? I don't think this is just an Operating thing.


> I didn't implement this logic for any other conditions except for 'Crosses End'. However, ComparisonEnd is used for : the 'Inside' condition (ComparisonEnd > Sys.Date or ExitAdjust <= ComparisonEnd) and (EntryDate >= ComparisonStart), the 'After' condition (EntryDate > ComparisonEnd), the 'Crossed Period' condition (ExitAdjust > ComparisonEnd and EntryDate < ComparisonStart),
> 
> Should I add the condition : !is.na(ComparisonEnd) & ComparisonEnd >= Export Start to each? I think it makes sense for 'After' and 'Crosses Period' but I am unsure about 'Inside'.

I guess the first question is what do we _want_ the outcome to be if ComparisonEnd is NA or < ExportStart? Right now, if it's NA, then Enrollmentvs returns NA. But I think it ought to return "Inside". I'm not sure why you added the >= ExportStart condition, nor what we'd want to happen if if it's < ExportStart.

> Which are removed in another branch? Are the conditions removed from Enrollmentvs or just the eva checks on them?

I think I'm thinking about something else, but what I *had* meant was that the "Enrollment Crosses ... Period" checks seemed redundant. But I realize now that's not true. However, I do think "Enrollment Crosses ... Period" is a bit of a misnomer; it's more precisely that the enrollment contains the entire period. But let's leave as-is.

alex-silverman:
> @alex-silverman Note that the shinytest changes were merged from 1061 into dev. See this commit: [dcb484a](https://github.com/abtassociates/eva/commit/dcb484ae4e23e793910117ffc49319e3b2468fda)
> 
> I think when you use the 'resolve conflicts' functionality for the shiny tests, it ignores those differences with 'dev' when it runs the PR shiny tests (allowing the tests to pass on #1064). Then, when it was pulled into this branch, the shiny tests fail when it encounters the difference and does not fall back on the conflicts being "resolved"
> 
> I would have assumed the shiny tests for 1064 would have failed, but they did not.

I'm not sure I'm following. The conflicts in the shiny tests were not resolved in the snapshots that were part of the PR (https://github.com/abtassociates/eva/blob/1061-new-dq-check-hoh-exited-before-other-hh-member/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-dq-org.json#L1086-L1092). The conflicts would need to be resolved and committed on the working branch for the tests to pass)

alex-silverman:
> > @alex-silverman Note that the shinytest changes were merged from 1061 into dev. See this commit: [dcb484a](https://github.com/abtassociates/eva/commit/dcb484ae4e23e793910117ffc49319e3b2468fda)
> > I think when you use the 'resolve conflicts' functionality for the shiny tests, it ignores those differences with 'dev' when it runs the PR shiny tests (allowing the tests to pass on #1064). Then, when it was pulled into this branch, the shiny tests fail when it encounters the difference and does not fall back on the conflicts being "resolved"
> > I would have assumed the shiny tests for 1064 would have failed, but they did not.
> 
> I'm not sure I'm following. The conflicts in the shiny tests were not resolved in the snapshots that were part of the PR (https://github.com/abtassociates/eva/blob/1061-new-dq-check-hoh-exited-before-other-hh-member/tests/testthat/_snaps/linux-4.5/main-valid/test-main-valid-dq-org.json#L1086-L1092). The conflicts would need to be resolved and committed on the working branch for the tests to pass)

I think I see what happened. I think the last merge to dev hadn't had the shinytest conflicts resolved yet. Just force-pushed resolutions to dev.

alex-silverman (approved):
These changes look good. We still need to revisit the EnrollmentVs function at some point, but this solves the immediate issue

PR #1063, opened 2026-08-26
in 05, `health_insurance_subs` need to replace NA values with 0s

alex-silverman:
For the individual health insurance cols, need to replace NA with 0 so that when we create `SourceCount`, it won't be set to NA if any of them are NA (even if some are 1). We were undercounting conflicting records because NAs were ignored.

I also added some code to get a better mirai traceback if there's an error.

### DIFF
--- a/rcode/05_data_quality.R
+++ b/rcode/05_data_quality.R
@@ -1411,7 +1411,7 @@ health_insurance_subs <- base_dq_data_inc %>%
   ) %>%
   replace_na(value = 0, cols = hi_cols) %>%
   fmutate(
-    SourceCount = reduce(hi_cols, `+`, na.rm = TRUE)
+    SourceCount = Reduce(`+`, gv(., hi_cols))
   ) %>%
   fsubset((InsuranceFromAnySource == 1 &
             SourceCount == 0) |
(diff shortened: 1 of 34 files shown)


rcode/05_data_quality.R
```diff
@@ -1389,26 +1389,29 @@ missing_health_insurance_exit <- missing_health_insurance %>%
...
-    'IndianHealthServices',
-    'OtherInsurance'
+    hi_cols
   ) %>%
+  replace_na(value = 0, cols = hi_cols) %>%
   fmutate(
-    SourceCount = Medicaid + SCHIP + VHAServices + EmployerProvided +
-      COBRA + PrivatePay + StateHealthIns + IndianHealthServices +
-      OtherInsurance + Medicare
+    SourceCount = reduce(hi_cols, `+`, na.rm = TRUE)
```
mlfurman3: I think we need to use `sum` here in order to pass additional args, otherwise it gives an error for unused argument. If that doesn't work, may need `across` or something similar


```suggestion
    SourceCount = reduce(hi_cols, sum, na.rm = TRUE)
```

alex-silverman: should be fixed!

mlfurman3 (approved):
This looks good to go now, thanks!

PR #1064, opened 2026-08-27
1061 new dq check hoh exited before other hh member

sarasoko:
# EvaChecks.csv
- Add new High Priority dq check
- Reorder dq 'Warning' and 'Error' from end of script to be with other dq checks

# 05_data_quality.R
- add new check 'hh_hoh_exit' for HOH exiting before other household members
- rowbind 'hh_hoh_exit' in 'hh_issues' and remove it

# server_00_changelog.R
- Add change log entry

# update shiny tests


rcode/05_data_quality.R
```diff
@@ -231,10 +231,25 @@ hh_missing_rel_to_hoh <- base_dq_data %>%
   merge_check_info_dt(checkIDs = 4) %>%
   fselect(vars_we_want)
 
+hh_hoh_exit <- base_dq_data %>% 
+  fgroup_by(HouseholdID) %>%
+  fmutate(
+    nPeople = length(unique(PersonalID)), # GRPN()?
+    nExits = length(unique(ExitDate)),
```
mlfurman3: Thanks @sarasoko, I would try `fnunique` here fpr speed but otherwise looks good

```suggestion
    nPeople = fnunique(PersonalID), 
    nExits = fnunique(ExitDate),
```

rcode/05_data_quality.R
```diff
@@ -231,10 +231,25 @@ hh_missing_rel_to_hoh <- base_dq_data %>%
   merge_check_info_dt(checkIDs = 4) %>%
   fselect(vars_we_want)
 
+hh_hoh_exit <- base_dq_data %>% 
+  fgroup_by(HouseholdID) %>%
+  fmutate(
+    nPeople = fnunique(PersonalID), 
+    nExits = fnunique(ExitDate),
+    #earliestExitDate=fmin(ExitDate, na.rm=FALSE),
+    lastExitDate=fmax(ExitDate, na.rm=FALSE)
```
alex-silverman: do you want `na.rm=TRUE` here? Otherwise, it will return NA if there are any NAs, which would have downstream effects in the `fsubset`.

alex-silverman (commented):
one small comment on NA handling in the `fmax` but otherwise looks great! Nicely done.

rcode/05_data_quality.R
sarasoko: Nice Catch! I definitely meant to use TRUE

public-resources/EvaChecks.csv
```diff
@@ -31,6 +31,7 @@ dq,High Priority,Duplicate Entries,"3.10 - Project Start Date, 3.11 - Project Ex
...
 dq,High Priority,No Head of Household,3.15 - Relationship to Head of Household,This household does not have a designated head of household. Every household must have exactly one head of household. Please be sure: 1. All members of the household are included in the program stay; 2. Each household member's birthdate is correct; 3. Every household member has a Relationship to Head of Household; 4. Exactly one member is designated as the head of household for the entire duration of the enrollment.,,2
 dq,High Priority,Too Many Heads of Household,3.15 - Relationship to Head of Household,"This household has more than one head of household. According to the HMIS Data Standards Manual, one member must be designated as the head of household for the entire duration of the enrollment.",,3
+dq,High Priority,Head of Household Exited Before Other Household Members,3.15 - Relationship to Head of Household,"The Head of Household (HoH) for this enrollment was exited earlier than the other household members. A HoH must be designated for the entire duration of the enrollment. If the Exit Date for this client is accurate, another household member in the enrollment must be designated as the HoH.",,24
```
alex-silverman: where did we land on using Check ID numbers that were likely once used by a different check? Do we want to make this a higher number, like 145 (which is the max + 1)?

rcode/server_00_changelog.R
alex-silverman, line 1: let's update to use the new MD approach

public-resources/EvaChecks.csv
```diff
@@ -77,6 +78,7 @@ dq,Error,Oldest Household Member Under 12,"3.03 - Date of Birth, 5.09 Household
...
 dq,Error,No bed night on Entry Date,"3.10 - Project Start Date, 4.14 - Bed-Night Date",This ES - NbN enrollment is missing a bed night record that matches the enrollment's 'Entry Date'. There must be a record of a bed night on the 'Entry Date' into a night-by-night shelter; any additional bed night dates must be after the 'Entry Date' and before the 'Project Exit Date'.,,107
 dq,Error,Bed night entered on Project Exit Date,"3.11 - Project Exit Date, 4.14 - Bed-Night Date",This ES - NbN enrollment has a bed night record that matches the enrollment's 'Project Exit Date'. Bed night dates must be after the 'Project Start Date' and before the 'Project Exit Date''.,,108
+dq,Error,Enrollment During Non-Participation Period,"2.08.1 HMIS Participation Type, 2.08.2 HMIS Participating Start, 2.08.3 HMIS Participating End","This enrollment is within a period of time when the project was not participating in HMIS. For reporting to work correctly, please be sure your enrollment dates are within the project's HMIS participation dates. If this project is HMIS Participating, the participation records may need to be updated.",,144
```
alex-silverman, line 81: good catch reordering these

alex-silverman (commented):
one comment on check ID and just need to fix changelog Otherwise, looks good!

public-resources/EvaChecks.csv
sarasoko: That's a good point! I have been trying to do the opposite. Choosing the smallest unused number. I can make this change

rcode/server_00_changelog.R
sarasoko: merged changes from dev

alex-silverman (approved):
once the tests pass, good to merge!

alex-silverman:
@sarasoko - looks like tests didnt pass. Can you double check those?

PR #1065, opened 2026-08-31
1051 speed up mirai serialization

alex-silverman:
### Primary Purpose
- Reduce memory usage to accommodate very large (200MB+) uploads and avoid crashes.

### Key Changes
- Reducing memory usage
  - Using `qs2` package to serialize the DQ+PDDE dependencies and save them to .qs2 file in a user-session-specific temp folder before calling the mirai, then passing the mirai the filepath to the .qs2 file. This substantially reduces memory and slightly speeds up the set up of the mirai because we're only passing a filepath.
  - subsetting IncomeBenefits and HealthAndDV earlier on
  - removing Enrollment dataset copies
  - removing intermediate datasets after use throughout 04 and in the mirai.
- Tracking memory use throughout processing.
- Sandbox script to create a synthetic massive datafile for testing
- Sandbox script to better compare different mirai approaches. Likely no longer necessary having identified `qs2` as a good solution, but may want to come back to it at some point.


mlfurman3:
Tested this out a bit, works well and am ok to merge in once conflicts are resolved

mlfurman3 (approved):

PR #1066, opened 2026-08-31
Code cleanups

alex-silverman:
### Primary Purpose
- Reduce repetitive code and improve code organization

### Key Changes
- Combined and simplified code into reusable functions for cleaner code, primarily in System scripts
- Renamed outputs, datasets, and variables for consistency.
- moved code related to specific tabs into their tab-specific `server_` script for code organization
- Split up CSS into separate, component/tab-specific scripts


mlfurman3 (commented):
@silvermana6  Generally looking really good, but want to keep/restore the exits by subpop selector checkboxes

PR #1069, opened 2026-09-01
validate not enough data for comp chart after suppression

alex-silverman:
validate not enough data for comp chart after suppression. Fixes the following log error:
Warning: Error in scale_fill_gradient: Discrete values supplied to continuous scale.
â„¹ Example values: NA, NA, NA, NA, and NA


mlfurman3 (approved):
Definitely a needed fix, thanks for implementing!

PR #1070, opened 2026-09-01
fixing log error:

alex-silverman:
[SESSION ID] 2026-08-31 14:07:45.593296  In get_active_info Warning: Error in map2: â„¹ In index: 1.
â„¹ With name: Permanent.
Caused by error in `if (count_prev == 0) ...`:


fallenmi (approved):
The zero-length guard fixes the concrete R failure here: when the previous-period summary is empty, count_prev == 0 produces logical(0), which cannot be used as the condition of a scalar if. Returning the same sentinel already used for a numeric zero preserves both call modes ("-" for table/export output and NA for numeric chart output), while leaving every nonempty calculation unchanged.

Reviewed with OpenAI Codex and Anthropic Claude against exact head 7a1ed3d676c1518e19d7c50ec659976935e95129. The current public shiny-test-automation check, check suite, and workflow run are green; I did not run tests locally.

mlfurman3 (approved):
Looks good, ok to merge in

PR #1082, opened 2026-09-18
Hotfix/missing inc data

silvermana6:
Fixes an issue with the Missing Income DQ checks which was due to a join that had been switched in the last push. Originally we left-multi-joined base_dq_data to IncomeBenefits, then later subsetted DataCollectionStage == 1. In this way, we ensure both Enrollment- and IncomeBenefits-level filters were applied.

In the last push, we subsetted each dataset first before joining, but we also flipped the join. Thus some IncomeBenefits records that did not meet the enrollment filter but met the IncomeBenefits filter remained, leading to empty enrollment-level fields (e.g. ProjectName and OrganizationName) in the export.


mlfurman3 (approved):
Looks good, thanks for the quick fix!

PR #1083, opened 2026-09-18
1061 new dq check hoh exited before other hh member

sarasoko:
Correction addressing @vlopez0603 comments concerning cases where the HOH is the only Exit (all other household members active) 

Rather than modifying the logic significantly, I swapped ExitDate for ExitAdjust when making and comparing lastExitDate so that the na.rm does not drop any dates.


changelog.md
```diff
@@ -1,5 +1,10 @@
 
 # Eva Changelog
+## 2026-09-24
```
silvermana6:
```suggestion
## 2026-10-01
```
Pushes are typically every 2 weeks

silvermana6 (commented):
I'm a little confused on this one. Why does using ExitAdjust - which is the same as ExitDate, except that NAs are replaced with 2099-09-09 - account for cases where the HoH is the only exited household member for an enrollment?

PR #1085, opened 2026-09-18
Hotfix/limit lh dates

silvermana6:
This attempts to resolve the 32,727 character Excel-export issue. We suspect that by including all LH dates for a given enrollment, even those more than 3 years in the past, we are passing this limit when those dates are concatenated for the export. Limiting to just the last 3 years ensures we have no more than 12 characters per date ("YYYY-MM-DD, ") * 365 dates per year * 3 years = 13,140 characters.


mlfurman3 (approved):
Makes sense and seems to work as expected - I tested with the `demo_lots_of_lh_dates.zip` upload

